PR #50, opened 2015-01-10
Added support for modal boxes Added feature for super admin to be able to edit user details. Using Modal box for edit feature.

puneet-sutar:


puneet-sutar:
http://wheel-add-supprt-for-modal-box.herokuapp.com/

puneet-sutar:
@neerajdotname Doing this task made me realise it would be good if we can display the validation errors within the modal box. For now we just redirect and show errors on the main page http://monosnap.com/image/JhPuRXtngCS2dt7c6DiTmKpiVcwfS0

neerajdotname:
Validation message has to be on the modal. Otherwise the user experience is terrible.

puneet-sutar:
@neerajdotname Can you verify now. If you approve of this approach then we should we change the code for 7Atlis also ?

neerajdotname:
@puneet-sutar looks good. And the approach is good. In 7atlis so far we have not had a form so this is not an issue right now but yes we should do the same there too.

PR #55, opened 2015-01-20
use mail_interceptor

neerajdotname:


vipulnsward:
Need to remove https://github.com/bigbinary/wheel/blob/master/config/initializers/email_interceptor.rb

PR #56, opened 2015-02-01
Content Compression via Rack::Deflater

justanshulsharma:
- [x] Content Compression via Rack::Deflater
for more info check http://robots.thoughtbot.com/content-compression-with-rack-deflater


justanshulsharma:
@neerajdotname I think we should use Rack::Deflater

PR #58, opened 2015-03-01
 Fix secret_key for production environment

prathamesh-sonpatki:
- Added the secret key in the ERB tags so that its evaluated as Ruby code.


prathamesh-sonpatki:
@vipulnsward Please check.

PR #65, opened 2015-04-05
Remove upgrade warning 

rkr090:
`rake setup` is showing following warning.

```
UPGRADE WARNING: Honeybadger.configure was removed in v2.0 and has no effect. Please upgrade: https://www.honeybadger.io/s/gem-upgrade
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -132,7 +132,7 @@ GEM
       activesupport (>= 3.2, < 5)
     hashr (0.0.22)
     hike (1.2.3)
-    honeybadger (2.0.6)
+    honeybadger (2.0.11)
     html2haml (2.0.0)
       erubis (~> 2.7.0)
       haml (~> 4.0.0)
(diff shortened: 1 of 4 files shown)


rkr090:
The following line could have been moved to honeybadger.yml, but in the name of consistency i have kept it in settings.yml .

```
<%= ENV['HONEYBADGER_API_KEY'] %>
```

This means that 'bundle exec honeybadger test' or any other honeybadger command will not work as it needs rails to be loaded to access 'Settings' config. (honeybadger comes with its own binstub)

So, i have added a binstub for honeybadger in the bin directory, every honeybadger command have to be executed through './bin/honeybadger', this way it will have access to the 'Settings' config.

```
./bin/honeybadger test
```

I couldn't do thorough testing as i don't have a project with delayed job and honeybadger test account  setup, i plan to do that tomorrow.

rkr090:
@neerajdotname 

I have tested the PR on a sample app with honeybadger and delayed job configured.

Look good.

PR #69, opened 2015-05-08
Use filter in minitest.

mechanicles:


mechanicles:
Before filter,
http://take.ms/TGRi0

After filter,
http://take.ms/xCKgb

prathamesh-sonpatki:
Looks good :+1:

PR #70, opened 2015-05-11
do not use branch name in db name

neerajdotname:


config/database.yml.postgresqlapp
```diff
@@ -4,12 +4,15 @@
 # Check doc/why_database_name_only_63_characters_long.md to see
 # details about the restriction on database name
 
+# If you prefer to have branch name in the database name then use following
+# database: <%= "#{repository_name}_development_#{branch_name}"[0...63] %>
+
 development:
   adapter: postgresql
-  database: <%= "#{repository_name}_development_#{branch_name}"[0...63] %>
+  database: <%= "#{repository_name}_development"[0...63] %>
```
vipulnsward: Need to remove `[0..63]`

vipulnsward:
:+1: I end up removing this in every project.

config/database.yml.postgresqlapp
mechanicles: Yup. I also don't use it. I keep it simple.

neerajdotname: I would suggest to keep it. It does not do any harm and if the repo name is long then it fixes thing.

vipulnsward: @neerajdotname yes, realised that. Its merged now.

PR #72, opened 2015-05-15
 Add daemons gem to Gemfile

prathamesh-sonpatki:
- Delayed job requires daemon gem if we want to run it as background
   process.
 - Thin gem also depends on daemons gem so it was working for a long
   time without explicitly requiring it.
 - But we moved thin in development and now it doesn't work in
   non-development environemnts.
 - Reference: https://github.com/collectiveidea/delayed_job#running-jobs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,6 +73,9 @@ gem 'mail_interceptor'
 # not used in development.
 gem 'unicorn', group: [:staging, :production]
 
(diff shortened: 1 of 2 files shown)


prathamesh-sonpatki:
@nishantmodak Found this problem in Emami.
//cc @neerajdotname

PR #73, opened 2015-05-15
Use secrets.yml instead of settings.yml

vipulnsward:


vipulnsward:
@neerajdotname @prathamesh-sonpatki  Please review.

Gemfile
```diff
@@ -32,9 +32,6 @@ gem 'delayed_job_active_record'
 # web interface for delayed job
 gem 'delayed_job_web', '>= 1.2.0'
 
-# collection of handy tools
-gem 'handy'
-
```
prathamesh-sonpatki: Do we have to remove handy completely?

config/secrets.yml
```diff
@@ -10,16 +10,91 @@
 # Make sure the secrets in this file are kept private
 # if you're sharing your code publicly.
 
+default: &default
```
prathamesh-sonpatki: Can be made `common` like we had in settings.

Gemfile
vipulnsward: Anywhere else, we would need this? cc @neerajdotname

config/secrets.yml
vipulnsward: `default` is more apt, and maps better to rails way of using `default`, like in the case of `database.yml`

Gemfile
prathamesh-sonpatki: We have some useful tasks here - https://github.com/bigbinary/handy/blob/master/lib/handy/additional_test_directories.rb

vipulnsward: Ahh, right

Vipul A.M.
+91-8149-204995

config/secrets.yml
prathamesh-sonpatki: `common` implies sharing common code like what is happening here. I am fine with `default` too :smile:

config/secrets.yml
```diff
@@ -10,16 +10,91 @@
...
+      authentication: plain
+  devise:
+    secret_key: c0cd8c380935936ff6c6bfb7b0f0ed3f3a2d66d7d7b445394a879dea71bfb5711799c7840e7d4fd93fe8a0983c4f27fd30caebea879dff734c15ff1696262144
 
 test:
-  secret_key_base: b731f261415e6f13e3e3e5a6726ee163707938f38514331fc2a67eda3800dd428bfe2a0c3b4ed83cf99e7df643e93b450511f81e05556712128ae786559ada6e
+  <<: *default
+  secret_key_base: 08523cd7820d8e6232afb3a704c60a75feb024ab937de989322f4e64a3854b4d8dc8bb119faf8847adceb70aadce1288ff3193ce62de10f78da223e95760dd6c
+  mailer_delivery_method: :test
+  mailer:
```
prathamesh-sonpatki: I guess `mailer` settings can be extracted in `common/default`?

vipulnsward: We assign this as a hash to smtp settings directly, so cant extract it, else will need to manually do hash merge, etc.

prathamesh-sonpatki: We assign only `smtp_settings` which can be common. `delivery_method` can be overridden in specific environments like we had previously in settings.yml

vipulnsward: It can differ, compare across this file. Some use auth, some use starttls, etc. Some don't.
For testing as well as development, I find this handy, because we don't need to copy over the settings to test out delivery.

config/honeybadger.yml
```diff
@@ -1,2 +1,2 @@
 ---
-api_key: <%= Settings.honeybadger_api_key %>
+api_key: <%= ENV['HONEYBADGER_API_KEY'] %>
```
neerajdotname: This needs to be reverted. I do not want to look into initializer to see the values. That is why settings.yml was there. And same needs to be there in secretys.yml too.

PR #83, opened 2015-05-28
 Added sample eslintrc file with ES6, JSX support enabled

prathamesh-sonpatki:


mechanicles:
@prathamesh-sonpatki But how it will help? Because I'm using Vim plugin for that.

prathamesh-sonpatki:
This is a sample global file. All of us can use this so that our configuration is consistent. Vim plugin will use configuration from this file. Consider this similar to .gemrc or .railsrc files. 

I added more details here - http://how-we-work.bigbinary.com/conventionandstyle/using-eslint.html

mechanicles:
:+1:

PR #84, opened 2015-05-28
Add https://github.com/basecamp/marginalia

prathamesh-sonpatki:


vipulnsward:
I am :-1: on this. When searching for logs for issues, first thing I do is search based on controller, then error, etc. 
This will add to effort when searching for issues. We don't have a better mechanism to search logs right now.
Finding hotspots in queries can be a task offloaded on database log, instead of AR log.

prathamesh-sonpatki:
@neerajdotname Please check 
84_wheel

neerajdotname:
Looks awesome.

PR #86, opened 2015-05-29
allow function names starting with underscore

neerajdotname:
statements like this._executeQuery is being flagged. This
PR would make such warnings go away.


prathamesh-sonpatki:
:+1:

PR #91, opened 2015-06-23
eslintrc should warn if "if statement" is not properly formatted

prathamesh-sonpatki:
```
var name = "John";

// good
if ( name === 'mary' ) {
 console.log("it is Mary");
} else if (name === 'Susan') {
 console.log('it is Susan');
} else {
 console.log("it must be John")
}


// bad
if( name === 'mary' ){
 console.log("it is Mary");
}else if(name === 'Susan'){
 console.log('it is Susan');
}else{
 console.log("it must be John")
}

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -38,6 +38,10 @@
     //In ReactNavite code it is common to have styles.xxx at the top while styles is defined later. PR #85
     "no-use-before-define": [0, "nofunc"],
 
-    "comma-dangle": "never"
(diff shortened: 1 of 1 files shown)


prathamesh-sonpatki:
users prathamesh projects fun eslint js 2015-06-23 20-15-33

@neerajdotname Please check

prathamesh-sonpatki:
Also added http://eslint.org/docs/rules/space-before-blocks.html

PR #93, opened 2015-06-24
Fix comma-dangle rule.

srih4ri:


neerajdotname:
I'm not clear. Is dangling of comma allowed or not allowed.

If it is not allowed then I want it to be reversed. It must be allowed. Otherwise it is such a pain to add one more item. And if someone leaves a dangling comma it's no big deal for me.

srih4ri:
@neerajdotname With `comma-dangle` set to `never` - It warns whenever a trailing comma is detected.
Other options are `always` and `always-multiline` .

From http://eslint.org/docs/rules/comma-dangle:

If "always-multiline" then it warns if there is a missing trailing comma on arrays or objects that span multiple lines, and warns if there is a trailing comma present on single line arrays or objects. If "never" then it warns whenever an trailing comma is detected. The default value of this option is "never".

The docs also say: 

You can turn this rule off if you are not concerned with dangling commas.

Should this rule be turned off?

neerajdotname:
Too much warning. Just turn this check off. 

People should be allowed to

* have dangling comma
* not have dangling comma

No need to warn. 

Basically company is not enforcing any rule.

prathamesh-sonpatki:
My bad. Dangling comma should be allowed.

PR #94, opened 2015-06-24
Remove comma dangle rule.

srih4ri:
See https://github.com/bigbinary/wheel/pull/93#issuecomment-114876329


prathamesh-sonpatki:
@srih4ri I think it should be `"comma-dangle": [0]` can you check?

srih4ri:
@prathamesh-sonpatki Yes [it should be](http://eslint.org/docs/user-guide/configuring.html#configuring-rules) . Changing.

PR #96, opened 2015-06-25
 Don't show warnings when camelcase variable names are used

prathamesh-sonpatki:
- Sometimes we have to use camel case variable names especially for
   sending data to API.


prathamesh-sonpatki:
@neerajdotname Please check.

PR #101, opened 2015-08-06
Adding fog gem and carrierwave initializer

ershad:


Gemfile
```diff
@@ -55,6 +55,8 @@ gem 'activeadmin', github: 'activeadmin'
 
 # for handling file uploads
 gem 'carrierwave'
+gem 'fog', require: false
```
neerajdotname: Please add proper comment why 'fog' is needed. I think it's because carrierwave uses 'fog' to  handle s3 uploads. If that's the case then mention that.

Gemfile
```diff
@@ -55,6 +55,8 @@ gem 'activeadmin', github: 'activeadmin'
...
+gem 'mini_magick'
```
neerajdotname: I think this is needed for image manipulation. If I do not want image manipulation then I think I can comment it out. Right ?

ershad: Yes. I've commented it out for now. It could be enabled when needed.

ershad:
@neerajdotname updated the PR, added comments in Gemfile. Please review.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,39 @@
...
+    CarrierWave.configure do |config|
+      config.root              = Rails.root.join('tmp')
+      config.storage           = :file
+      config.enable_processing = false
+    end
+  else
+
+    if Settings.store_images_in_file_system
+
+      CarrierWave.configure do |config|
```
neerajdotname: Can we extract the whole block into a method called `setup_for_storing_images_in_file_system ?

I'm find it hard to follow logic with so much code inside a single case statement.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,39 @@
...
+        config.permissions           = 0600
+        config.directory_permissions = 0700
+        config.storage               = :file
+        config.asset_host            = Settings.url_with_port
+      end
+
+    else
+
+      CarrierWave.configure do |config|
```
neerajdotname: Can we extract the whole block into a method called `setup_for_storing_images_in_s3 ?

I'm find it hard to follow logic with so much code inside a single case statement.

ershad:
@neerajdotname extracted those lines of code to methods.

config/settings.yml
```diff
@@ -0,0 +1,67 @@
...
+      password: <%=ENV['SENDGRID_PASSWORD']%>
+      domain: <%=ENV['SMTP_DOMAIN']%>
+      address: smtp.sendgrid.net
+      port: 587
+      authentication: plain
+      enable_starttls_auto: true
+
+  host: <%= ENV['HOST'] %>
+  url_with_port: <%= ENV['HOST'] %>
+  host_for_email_with_www: <%= ENV['HOST'] %>
```
vipulnsward: @ershad We have moved Settings.yml to Rails application secrets, can you move/merge there?

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,42 @@
...
+    config.asset_host            = Settings.url_with_port
+  end
+end
+
+def setup_for_storing_assets_in_s3
+  CarrierWave.configure do |config|
+    config.root            = Rails.root.join('tmp')
+    config.fog_credentials = {
+        provider:              'AWS', # required
+        aws_access_key_id:     Settings.aws_access_key_id, # required
```
vipulnsward: Please move and use Rails.application.secrets everywhere instead of Settings

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,42 @@
...
+        aws_secret_access_key: Settings.aws_secret_access_key, # required
+        region:                'us-west-2' # required
```
neerajdotname: Remove the comments like `required`. That's noise and does not add any meaningful value to code.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,42 @@
...
+    }
+    config.fog_directory   = Settings.s3_bucket
+    config.fog_public      = true # optional, defaults to true
+    config.fog_attributes  = { 'Cache-Control' => 'max-age=315576000' } # optional, defaults to {}
```
neerajdotname: Please tell me what value the comment provides. It's nothing but noise to me.

ershad:
@vipulnsward @neerajdotname made the changes, please review.

config/secrets.yml
```diff
@@ -80,6 +82,7 @@ staging:
 
 production:
   <<: *default
+  store_assets_in_file_system: false
```
neerajdotname: Here the value is `false` which is right. But it does not make it explicit that the alternative is storing in s3. There are alternatives to s3 too. 

I think a better key would be `store_uploaded_assets_in` with correct values being either `file_system` or `s3`.

ershad:
@neerajdotname done.

Also please add me to Wheel in FreshBooks.

config/initializers/carrierwave.rb
```diff
@@ -0,0 +1,42 @@
...
+  when 'test'
+    CarrierWave.configure do |config|
+      config.root              = Rails.root.join('tmp')
+      config.storage           = :file
+      config.enable_processing = false
+    end
+  else
+    if Rails.application.secrets.store_uploaded_assets_in == 'aws_s3'
+      setup_for_storing_assets_in_s3
+    else
```
neerajdotname: Please explictly check for 'filesystem`. If there is a typo in `aws_s3` then it will start using `filesystem`. And add an else clause if there is no match and raise exception to catch typo scenarios.

ershad:
@neerajdotname made the change, it now explicitly checks for ```store_uploaded_assets_in``` value. If not found it would raise an exception. Please review.

neerajdotname:
Looks good. Please squash all the commits.

ershad:
@neerajdotname squashed the commits.

PR #110, opened 2015-11-20
Enable logging from dj

neerajdotname:
Currently `Rails.logger` message disappear in thin area. This will help see those logs.

It has been tested in Digital Ocean as well as heroku.

/cc @rkr090 @ershad @vipulnsward

### DIFF
--- a/config/initializers/delayed_job_config.rb
+++ b/config/initializers/delayed_job_config.rb
@@ -10,6 +10,17 @@ Delayed::Worker.read_ahead          = 10
 
 Delayed::Worker.delay_jobs          = !Rails.env.test?
 
+if ENV['HEROKU'].present?
+  Delayed::Worker.logger = Logger.new(STDOUT)
+else
+  Delayed::Worker.logger = Logger.new(File.join(Rails.root, 'log', 'delayed_job.log'))
+end
+
+if Delayed::Worker.delay_jobs && ($PROGRAM_NAME =~ /delayed_job/)
+  ActiveRecord::Base.logger = Delayed::Worker.logger
+  Rails.logger = Delayed::Worker.logger
+end
+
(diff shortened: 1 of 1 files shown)


config/initializers/delayed_job_config.rb
```diff
@@ -10,6 +10,17 @@
...
 
+if ENV['HEROKU'].present?
+  Delayed::Worker.logger = Logger.new(STDOUT)
+else
+  Delayed::Worker.logger = Logger.new(File.join(Rails.root, 'log', 'delayed_job.log'))
+end
+
+if Delayed::Worker.delay_jobs && ($PROGRAM_NAME =~ /delayed_job/)
+  ActiveRecord::Base.logger = Delayed::Worker.logger
+  Rails.logger = Delayed::Worker.logger
```
vipulnsward: I dint get this. This will put all logs from `production.log` to `delayed_job.log` for anything run from `delayed_job` workers.
So that is two places where we would need to debug information. Is this expected?

neerajdotname: Right now `Rails.logger.info` message is not appearing in any BB application which is hosted anywhere if that code is executed by DJ. 

This is the only way I could make it work after two days trying various options and understanding how monit starts our DJ. 

If anyone can make it work in any other way then let's discuss.

neerajdotname: Also as per boxy setting nothing comes to `production.log`. It goes to `unicorn.stdout.log`. https://github.com/bigbinary/boxy/blob/master/unicorn/templates/config/stage.rb.erb#L41-L42

neerajdotname: In AAPC this code change is working. Now are able to see all logs created in dj in  `delayed_job.log` and I tested this feature in heroku. So going to merge it.

PR #119, opened 2016-03-31
Remove traces of unicorn

vipulnsward:
- [x] Remove config/unicorn.rb
- [x] Procfile is still referring to unicorn https://github.com/bigbinary/wheel/blob/85de97864cf86cd0ee96ff63ed7cd6adc6cc109b/Procfile

@vipulnsward _a


vipulnsward:
Deployed here- http://wheel-119-puma-config.herokuapp.com/

PR #122, opened 2016-08-16
Remove An orange ribbon at the top for non-production environment.

kalamkar:
- [x] Remove it from code
- [x] Update README

@kalamkar _a


kalamkar:
@neerajdotname _a Please review

PR #123, opened 2016-08-16
Enable turbo link

kalamkar:
- [x] in code
- [x] updated README

@kalamkar _a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,9 @@ gem 'sprockets-rails'
 gem 'sass-rails', '>= 5.0.3'
 gem 'uglifier', '>= 2.7.1'
 
+# Turbolinks makes navigating your web application faster. Read more: https://github.com/turbolinks/turbolinks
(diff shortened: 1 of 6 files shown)


kalamkar:
@neerajdotname _a Please review

neerajdotname:
@kalamkar _a Please keep javascript tag in the footer.

kalamkar:
@neerajdotname 

Turbolinks automatically fetches the page, swaps in its `<body>`, and merges its `<head>`

When we put `javascript_include_tag` in the footer, it reloads JavaScript files for every `<body>` swap and try to redefine already defined JavaScript modules.

As Turbolinks makes navigating web application faster, IMO it would not affect the performance when `javascript_include_tag` is in the `<head>`. 

Please let me know your views.

neerajdotname:
@kalamkar Good point. I had forgotten that now we have enabled turbo link.

PR #131, opened 2016-09-30
I guess handy is not needed anymore

neerajdotname:


prathamesh-sonpatki:
It was needed earlier only for running `test:workers` etc. No longer needed in Rails 5.

PR #135, opened 2016-10-14
switch from mandrill to mailgun as default email provider

ershad:
https://www.mailgun.com

@ershad _a


ershad:
@neerajdotname _a changed the configuration and tested mail delivery with Mailgun. Please review.

PR #137, opened 2016-11-16
Disable rack timeout in development

rkr090:
The output from rack-timeout is just too excessive and makes debugging hard.

### Before

before

### After

after

### DIFF
--- a/config/initializers/rack_timeout.rb
+++ b/config/initializers/rack_timeout.rb
@@ -1,1 +1,2 @@
+Rack::Timeout::Logger.disable if Rails.env.development?
 Rack::Timeout.service_timeout = 30 # seconds


config/initializers/rack_timeout.rb
```diff
@@ -1 +1,2 @@
+Rack::Timeout::Logger.disable if Rails.env.development?
```
vipulnsward: I would leave this as is. 
Production-development parity is important, breaking updates on rack-timeout should be caught on development.

neerajdotname: Let's add a comment there stating something to the effect of 

```
Following code is disabling Rack Timeout logger. Rack Timeout is enabled in development.
```
```

rkr090: I will add this comment to master.

PR #139, opened 2016-12-24
Use params[:email] instead of params[:id] for authentication

ershad:
In [authenticate_user_using_x_auth_token](https://github.com/bigbinary/wheel/blob/90832fe13be52a29fd121695396dc44c85cf4723/app/controllers/api/v1/base_controller.rb#L30) method, we are fetching the email using `params[:id]`. This could conflict with routes which have `id` param. Should we change it to `params[:email]`?

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -27,7 +27,7 @@ class Api::V1::BaseController < ApplicationController
   end
 
   def authenticate_user_using_x_auth_token
-    user_email = params[:id].presence
+    user_email = request.headers['X-Auth-Email']
     auth_token = request.headers['X-Auth-Token'].presence
 
     user = user_email && User.find_by_email(user_email)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -188,6 +188,7 @@ GEM
     jbuilder (2.6.0)
       activesupport (>= 3.0.0, < 5.1)
       multi_json (~> 1.2)
+    jquery-growl-rails (1.1.9)
     jquery-rails (4.1.1)
       rails-dom-testing (>= 1, < 3)
       railties (>= 4.2.0)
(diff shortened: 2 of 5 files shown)


ershad:

1. Refactored `Api::V1::UsersController`. Like any other resource, it should find the User object with `id` param. `email` param should be used only for authentication.
2. We can now pass `email` either in query string or in request header (`X-Auth-Email`). This way we do not have to pollute `params` object.

app/controllers/api/v1/base_controller.rb
```diff
@@ -27,7 +27,7 @@ def log_exception exception
   end
 
   def authenticate_user_using_x_auth_token
-    user_email = params[:id].presence
+    user_email = params[:email].presence || request.headers['X-Auth-Email']
```
sharang-d: @ershad _a Why should we support fetching the email from `params[:email]`? I think only headers would be a better way to go.

ershad: @sharang-d true, I would agree to that. But we have been passing email in query string for a while now, and I didn't know the reason behind, hence kept it like that.

@prathamesh-sonpatki @vipulnsward any reason why we should support passing email through params? If not, let's remove it.

sharang-d: @vipulnsward _a

vipulnsward: It exists for historical reasons and for APIs that don't deal with headers.

vipulnsward:
Please rebase 

@ershad _a

ershad:
@vipulnsward _a rebased. please review.

app/controllers/api/v1/base_controller.rb
sharang-d: @vipulnsward We are picking up the `auth_token` only from the headers so If an API call can be made with one header it can also be made with another header(for the email). What am I missing?

neerajdotname: @ershad let's do the right thing. Let's drop the support for `params[:email]`. Once it is done then ping me. We should merge it soon.

ershad:
@neerajdotname _a I've removed the support for params[:email] in authentication. Please review.

PR #141, opened 2017-01-09
Change controller tests to be the new style Rails 5 integration tests

sharang-d:
Change all test classes inheriting from `ActionController::TestCase` to inherit from `ActionDispatch::IntegrationTest` instead and make other changes as required.

Ref: https://youtu.be/_qG7A9LxBPw?t=738


sharang-d:
@prathamesh-sonpatki _a

prathamesh-sonpatki:
Nice 🤘

PR #142, opened 2017-01-18
Use entire heroku url from env variables

mohitnatoo:
@neerajdotname _a in Timely Radar I found that because of the existing code in the secrets.yml, the host in dev environment is taken by default as `https://.herokuapp.com`


prathamesh-sonpatki:
Looks good, can you check the build failure?

neerajdotname:
@mohitnatoo _a can you look into circle ci issue.

mohitnatoo:
@neerajdotname _a tests are green now.

PR #149, opened 2017-02-17
Add Trinity like notifications

rishi-mohan:
Add this notification functionality in wheel

image


rishi-mohan:
PR: 

* Added Growl Notification gem

SS: 

image

image

@neerajdotname _a

neerajdotname:
This is fine for now. Error messages should be sticky. Success messages should be growl type.

I will take care of this.

PR #152, opened 2017-02-22
Misc changes

neerajdotname:
Please see individual commit to see what is done here.


neerajdotname:
@vipulnsward @mohitnatoo Can anyone review it.

config/initializers/email_prefixer.rb
```diff
@@ -1,3 +1,3 @@
 EmailPrefixer.configure do |config|
-  config.stage_name = 'WHEEL'
+  config.stage_name = Rails.application.secrets.subject_prefix_for_outgoing_emails
```
mohitnatoo: `EmailPrefixer` has two config options
1. `application_name`.
2. `stage_name`

I think in the above case, I think we should be configuring `application_name`.
`stage_name` is to be used for the environment name (if we want to give any custom env name).

The final prefix combination is of the form `[application_name stage_name]`. So to have something like `[WHEEL Development]` or `[WHEEL Staging]`, we'll have `application_name` as WHEEL and then we can optionally provide `stage_name` if required. I think we should stick to that convention. What do you think ?

config/initializers/delayed_job_invoke_worker_automatically.rb
```diff
@@ -1,27 +0,0 @@
...
-  # $PROGRAM_NAME check is needed here to successfully run migrations
-  if !($PROGRAM_NAME =~ /(rake)(.rb)?$/) &&
-     ENV['INVOKE_WORKER_AUTOMATICALLY'] &&
-     ENV['INVOKE_WORKER_AUTOMATICALLY'] == 'enabled'
-
-    pid = spawn('INVOKE_WORKER_AUTOMATICALLY=disabled bundle exec rake jobs:work')
-    Process.detach(pid)
-  end
-
-end
```
vipulnsward: We do use this on heroku. Any reason to remove this?

vipulnsward: Actually, I think we don't and I don't remember the last time this was used so 👍

neerajdotname: @vipulnsward this was set this way by instapusher. And then you put anything in DJ and it would run. I don't think it was very reliable. Now with instapusher gone it is hard to configure it.

config/initializers/email_prefixer.rb
neerajdotname: @mohitnatoo I see. So to start with `stage_name` should not be `WHEEL`. 

I will keep the behavior same and will change `config.application_name` to the application name. 

Whether to pass env name or not can be taken up in another issue.

neerajdotname: @mohitnatoo done. I added env name also. If env name is `production` then this gem does not display env name which is what we want.

mohitnatoo:
LGTM.

@vipulnsward feel free to merge if you're fine with these.

PR #157, opened 2017-04-22
upgrade to ruby 2.4.1

neerajdotname:


neerajdotname:
Verified that deployed app on heroku is working. Merging it.

PR #160, opened 2017-04-22
Use published gem marginalia

neerajdotname:


neerajdotname:
Verified on heroku app. Merging it.

PR #147, opened 2017-06-09
Does anyone prefer slim over haml

neerajdotname:
I've always felt slim is much clean, readable and easy to work over haml. Does anyone else think wheel should switch to slim as they are very much similar and not much overload to switch.

Slim feels more html like, while haml is ruby like
screen shot 2017-02-04 at 20 03 38
screen shot 2017-02-04 at 20 04 18


neerajdotname:
Verified on heroku. Merging it.

PR #164, opened 2017-08-01
Add `prepend_before_action :authenticate_scope!` to resource based actions in `RegistrationController`

amit-choudhary:
We missed adding `prepend_before_action :authenticate_scope!` while adding `RegistrationsController` from https://github.com/plataformatec/devise/blob/v4.3.0/app/controllers/devise/registrations_controller.rb. Due to this while writing tests exception was raised. Explanation is below:
screen shot 2017-07-27 at 1 45 32 pm


amit-choudhary:

PR #173, opened 2017-12-11
Add daily cron

vipulnsward:
Since we don't do many changes on Wheel, added a cron to run everyday and verify things are ok.

This will report us issues like CVE, gem breakage, test failure, etc


neerajdotname:
Going forward please assign the PR to me and let's not use "review feature" of github.

vipulnsward:
Sorry, has become a habit, all other projects use this.

neerajdotname:
I understand.

PR #170, opened 2017-12-22
Update bootstrap to latest

rishimohan:


rishimohan:
Currently WIP

rishimohan:
@neerajdotname _a I've upgraded Bootstrap in the PR and fixed all the UI related issues, but there's one issue in which Edit User form in Superadmin Users page is causing "Malformed indentation" error and because of that, the Edit User modal isn't opening. I tried fixing it but none of the solutions worked. Will need help with this one. Here's the SS: 

image

neerajdotname:
@rishimohan We rarely use that feature. I have removed the code related to that.

PR #175, opened 2018-01-15
Upgrade and run rubocop -a

vipulnsward:


neerajdotname:
If anyone wants to catch up on `frozen_string_literal: true` then here are some relevant links.

https://bugs.ruby-lang.org/issues/8976

https://bugs.ruby-lang.org/issues/11473

https://twitter.com/yukihiro_matz/status/634386185507311616

PR #183, opened 2018-05-05
Add webpacker from master

vipulnsward:


rkr090:
webpack-dev-server version 3 is not compatible with webpack < 4.0. We will have to downgrade webpack-dev-server to 2.11.1

rkr090:
@vipulnsward _a please review

PR #220, opened 2019-04-01
Used Rails.application instead of Wheel::Application as per convention

chiraggshah:


chiraggshah:
@prathamesh-sonpatki _a : Please review.

PR #225, opened 2019-04-23
Update rubocop config to not update config files

prathamesh-sonpatki:


prathamesh-sonpatki:
Discussed with @rkr090 and looks good.

PR #257, opened 2019-10-14
Changed default ids to uuid

### LINKED ISSUE #200: Change default ids to uuid
https://blog.bigbinary.com/2016/04/04/rails-5-provides-application-config-to-use-uuid-as-primary-key

Since all our projects inherit wheel, this is becoming a chore. We should not use plain ids anymore in projects.

Rohann7:
Fixes #200


neerajdotname:
@Rohann7 Here is my feedback. https://www.loom.com/share/5ed1c250de9642929db2b574456190de

PR #279, opened 2020-01-20
By default heorku provisions pg version 11. We should use that in circleci for testing.

neerajdotname:


neerajdotname:
CircleCI test passed. Merging it.

PR #297, opened 2020-03-19
CircleCI builds fail fast for audit checks

danishsatkut:
I have moved bundle tasks before moving to any Node tasks. This allows a build to fail fast without downloading any node modules in case of bundle audit errors.

Please read https://github.com/bigbinary/nitrohelp-web/issues/3083 for more information.


danishsatkut:

PR #330, opened 2020-07-09
Remove rack-timeout

neerajdotname:


neerajdotname:
Verified it on heroku.

Issue #339, opened 2020-08-04
Add PurgeCSS


rishimohan:
About PurgeCSS: https://purgecss.com/

Using third-party libraries also adds tons of unused CSS/classes which doesn't get used. For ex. while using Tailwind we don't actually use all of the classes Tailwind provides, so compiled bundle will have all these unused classes which affects performance. To reduce the bundle size and to get rid of this unnecessary CSS we will add PurgeCSS.

This should be worked on after https://github.com/bigbinary/wheel/pull/338 is merged.

PR #345, opened 2020-08-13
Added CSS purging to reduce assets size on production

rishimohan:
Fixes https://github.com/bigbinary/wheel/issues/339

- Purging is not enabled in development

## Before purging: 

image

## After purging: 

image


rishimohan:
Merging, cc: @neerajdotname

PR #366, opened 2020-09-24
* Adds a index scaffold Template.

pankajdoharey:
* Adds a index page scaffold template as the default slim-rails doesnt work correctly.
* Replaces link_to 'Destroy' with button_to 'Destroy' as it doesnt work for post request.


pankajdoharey:
Closing incorrect branch name.

PR #415, opened 2020-10-21
Adds dotenv-rails to load env rails from .env file

unnitallman:
Related to https://github.com/bigbinary/nitrohelp-web/issues/7857

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,9 @@ source "https://rubygems.org"
 
 ruby "2.7.2"
 
+# To load environment variables from .env into ENV
+gem "dotenv-rails"
+
 gem "rails", "~> 6.0.3.4"
 gem "sprockets", "~> 3.7.2"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -101,6 +101,10 @@ GEM
       responders
       warden (~> 1.2.3)
     docile (1.3.2)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -4,6 +4,9 @@ source "https://rubygems.org"
 
 ruby "2.7.2"
 
+# To load environment variables from .env into ENV
+gem "dotenv-rails"
```
neerajdotname, line 8: @unnitallman Do we need this in all environments?

unnitallman: `development` and `test` will do usually since Heroku has its own setup for loading ENV variables. But this PR is related to running `staging` env on local - which means we need a way to load ENV variables from `.env` file on local machine as well. Hence added the gem without development/test scope.

neerajdotname: @unnitallman I see just the gem being added in the PR. Can you add instruction to run `staging` env on local or link to that.

unnitallman: That is done. That was the purpose of the issue for which I have sent this PR: https://github.com/bigbinary/neeto-help-web/issues/7857. Please see the issue description and the first comment.

neerajdotname:
Since we don't copy staging data to local very often this is best documented at https://how-we-work.bigbinary.com/. Having this gem in staging environment indeed seem confusing.

PR #428, opened 2020-10-25
Removes the unwanted server_rendering.js

unnitallman:


neerajdotname:
@unnitallman _a deployment is failing. Can you look into it.

PR #431, opened 2020-10-31
Added refactoring around migrations

sandip-mane:
- [x] Enabling the UUID in the first migration.
- [x] Setup `users.id` as `uuid` instead of `bigint`.
- [x] Removed unnecessary migrations.
- [x] Fixed an issue with the class_name in a test file.


sandip-mane:
@neerajdotname  _a please review.

PR #449, opened 2020-11-20
Add hero page

### LINKED ISSUE #436: Homepage should not take user to the login page
https://wheel-production.herokuapp.com/login should list a hero page with a link to login. In this way the app, by default, would be useful.

yedhink:
Fixes #436 

@vinay0x _a Would you be able to review this PR? Thanks.

cc: @edwinbbu 

Added a very simple hero page to stay aligned with the very simple login page, and also lists the features of `wheel` in that page.

Changes
- Added hero page
- Justified the button in login page to center as per Edwins suggestion
- Corrected spelling in Readme
- New user will be redirected to Hero page(path is `/`) before login
- User will be redirected to Hero page after logout
- Added ramda
- Modified the `PrivateRoute` to work with/without a `redirectRoute` - where without one and if `condtion` not met, then the `Hero` page gets rendered
- Use `getAttribute` for `Element` type value and provided default values as args

[Edit: Latest Image]
image

### DIFF
--- a/app/javascript/src/apis/axios.js
+++ b/app/javascript/src/apis/axios.js
@@ -3,11 +3,11 @@ import { Toastr } from "neetoui";
 
 axios.defaults.baseURL = "/";
 
-export const setAuthHeaders = (setLoading = () => null) => {
+export const setAuthHeaders = (setLoading = (e = null) => e) => {
   axios.defaults.headers = {
     Accept: "applicaion/json",
     "Content-Type": "application/json",
-    "X-CSRF-TOKEN": document.querySelector('[name="csrf-token"]').content,
+    "X-CSRF-TOKEN": document.querySelector('[name="csrf-token"]').getAttribute('content'),
   };
   const token = JSON.parse(localStorage.getItem("authToken"));
   const email = JSON.parse(localStorage.getItem("authEmail"));
--- a/app/javascript/src/apis/axios.js
+++ b/app/javascript/src/apis/axios.js
@@ -3,7 +3,7 @@ import { Toastr } from "neetoui";
 
 axios.defaults.baseURL = "/";
 
-export const setAuthHeaders = (setLoading = (e = null) => e) => {
+export const setAuthHeaders = (setLoading = () => null) => {
   axios.defaults.headers = {
     Accept: "applicaion/json",
     "Content-Type": "application/json",
--- a/app/javascript/src/components/Common/PrivateRoute.jsx
+++ b/app/javascript/src/components/Common/PrivateRoute.jsx
@@ -1,25 +1,35 @@
 import React from "react";
 import { Redirect, Route } from "react-router-dom";
 import PropTypes from "prop-types";
+import { either, isEmpty, isNil } from "ramda";
+
+import Hero from "components/Home/Hero";
 
 const PrivateRoute = ({
   component: Component,
   condition,
   path,
-  redirectRoute,
+  redirectRoute = null,
   ...props
 }) => {
-  if (!condition) {
+  if (condition) {
+    return <Route path={path} component={Component} {...props} />;
+  }
+
+  if (either(isNil, isEmpty)(redirectRoute)) {
+    return <Hero />;
+  } else {
     return (
       <Redirect
         to={{
           pathname: redirectRoute,
-          from: props.location,
+          state: {
+            from: props.location,
+          }
         }}
       />
     );
   }
-  return <Route path={path} component={Component} {...props} />;
 };
 
 PrivateRoute.propTypes = {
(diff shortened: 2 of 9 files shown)


app/javascript/src/apis/axios.js
```diff
@@ -3,11 +3,11 @@ import { Toastr } from "neetoui";
 
 axios.defaults.baseURL = "/";
 
-export const setAuthHeaders = (setLoading = () => null) => {
+export const setAuthHeaders = (setLoading = (e = null) => e) => {
```
vinay0x, line 6: Why do we need this change?

app/javascript/src/components/Common/PrivateRoute.jsx
```diff
@@ -1,25 +1,35 @@
...
   ...props
 }) => {
-  if (!condition) {
+  if (condition) {
+    return <Route path={path} component={Component} {...props} />;
+  }
+
+  if (either(isNil, isEmpty)(redirectRoute)) {
+    return <Hero />;
+  } else {
```
vinay0x, line 21: If user isn't authenticated redirect them to Hero page. Handle that in the main route file. PrivateRoute is a helper component. Custom logic shouldn't be handled in it. Please revert the PrivateRoute file.

app/javascript/src/components/Home/Hero.jsx
```diff
@@ -0,0 +1,46 @@
...
+        </p>
+      </div>
+      <ul className="list-disc ml-12 mt-6 text-gray-600 leading-relaxed tracking-wide">
+        {featureLists}
+      </ul>
+      <div className="mt-6 flex justify-center items-center">
+        <Button
+          className="ml-2"
+          type="primary"
+          onClick={() => { window.location.href = "/login" }}
```
vinay0x, line 40: Use `react-router` instead of `window.location`.
```suggestion
          to="/login" 
```

app/javascript/src/components/Home/Hero.jsx
```diff
@@ -0,0 +1,46 @@
...
+    "Content compression via Rack::Deflater.",
+    "Auto-formats Ruby code with rubocop.",
+    "Auto-formats JavaScript and CSS code with prettier.",
+    "Performs background job processing 'inline' for heroku env. It means heroku can deliver emails.",
+    "Letter opener gem for development.",
+  ];
+
+  const featureLists = features.map((feature, index) => (
+    <li key={index}>{feature}</li>
+  ));
```
vinay0x, line 21: Try to avoid using functions inside a component to render something. Use components to do it.

app/javascript/src/components/Home/Hero.jsx
```diff
@@ -0,0 +1,46 @@
...
+        <p className="max-w-xl mt-5 mx-auto text-xl leading-7 text-gray-800">
+          Don't reinvent the wheel. Use sane defaults to bootstrap your react-rails project!
+        </p>
+      </div>
+      <ul className="list-disc ml-12 mt-6 text-gray-600 leading-relaxed tracking-wide">
+        {featureLists}
+      </ul>
+      <div className="mt-6 flex justify-center items-center">
+        <Button
+          className="ml-2"
```
vinay0x, line 38: Remove this.

vinay0x (changes requested):
@yedhink _a Please check the comments.

app/javascript/src/apis/axios.js
yedhink: This is for automatic type acquiring in vscode.

app/javascript/src/components/Home/Hero.jsx
yedhink: Sure. I will make the change.

@vinay0x 
Just to clarify, what are the caveats associated with using a `map` function within a component to create jsx? Should I make it a general practice to extract out variables whose values are static and are composed of jsx elements?

app/javascript/src/apis/axios.js
vinay0x: It's not needed. Just use the shorthand no-op method.

app/javascript/src/components/Home/Hero.jsx
vinay0x: There are no caveats in using `map` within a component. My point was, don't have a method that renders something. Use another component. 

Either do 
```javascript
 features.map((feature, index) => (
    <Feature feature={feature}/>
  ));
```

or do 
```javascript
    <FeaturesList features={features}/>
```
and loop inside that component.

yedhink:
@vinay0x _a

I have made the requested changes. Let me know if we need to make anymore changes.

vinay0x (approved):

PR #457, opened 2020-12-01
Add jsconfig

### LINKED ISSUE #456: Add .jsconfig for wheel
@yedhink _a

yedhink:
Fixes #456 

@vinay0x
I will merge this minor addition.

### DIFF
--- a/app/javascript/src/jsconfig.json
+++ b/app/javascript/src/jsconfig.json
@@ -1,0 +1,27 @@
+{
+  "compilerOptions": {
+    "target": "ES6",
+    "allowSyntheticDefaultImports": true,
+    "strictNullChecks": true,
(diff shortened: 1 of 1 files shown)


vinay0x:
@yedhink Don't merge PRs unless all tests pass.

yedhink:
@vinay0x 
Thanks for pointing it out. I will definitely keep that in mind from next time.

The reason I went for the merge before the tests completed was because I was sure that the change doesn't affect the working of the application, since this is only a editor thingy. But now I realise the problem with merging before tests completion is that, the branch gets deleted once we merge(which I didn't know), thus leaving semaphore ci without any code base to test on, thus test fails.

Again, thanks for pointing this out. Learnt something new.

PR #465, opened 2021-03-03
Fixes an issue with git randomely ignoring files in namespace "tags"

sandip-mane:


sandip-mane:

//cc @yedhink

PR #473, opened 2021-04-12
Copy database.yml if not present

### LINKED ISSUE #472: Copy database.yml if not present while setup
Instead of raising an error, copy contents of `config/database.yml.postgresql` in `config/database.yml`.

This will be much helpful to the most of the people.

@datt _a

datt:
Fixes #472.


datt:
@neerajdotname _a Can we merge this? This would be much helpful for all.

PR #511, opened 2021-05-21
Add dotenv comptability

yedhink:
@josephv7 

Can you checkout this branch and see if it's what you had in mind? Please take a look into the changes in  README in this PR to understand workflow. Thanks.

@yedhink _a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,12 @@ gem "sidekiq"
 # Preview email in browser
 gem "letter_opener", group: :development
 
+# yanked v0.3.5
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -62,6 +62,12 @@ gem "sidekiq"
 # Preview email in browser
 gem "letter_opener", group: :development
 
+# yanked v0.3.5
+gem 'mimemagic', github: 'mimemagicrb/mimemagic', ref: '01f92d86d15d85cfd0f20dabd025dcbd36a8a60f'
```
akhil-gautam, line 66: Do we need `mimemagic` gem? Rails is already upgraded to `6.1.3.2`.

yedhink: @akhil-gautam Good catch. It's not needed. 

I'm closing this PR either ways, since it's not needed at the moment.

PR #525, opened 2021-06-04
Updated git hooks preventive message

### LINKED ISSUE #524: Not able to push repo to a new app
@yedhink _a Now that we do not allow push to master I was not able to push wheel to a new repo. We need to provide an override option. Something like `git push --no-verify` might do the work. If we adopt this then the error message should say that " ...... In order to override it please provide --no-verify option".

yedhink:
Fixes #524 

Screenshot_20210604_085044

@neerajdotname _a
Would you be able to give this a review and merge if everything looks good? Thank you.


neerajdotname:
@yedhink I see that the text has been changed but I see no code change. Has it been tested that passing `--no-verify` actually works?

yedhink:
@neerajdotname Yes. The `--no-verify` option from `git` itself can be leveraged for this task. I had tested committing and pushing to master in my own test repo with `git push --no-verify`, and it was working even if we had a hook to prevent from doing the same.

PR #531, opened 2021-07-06
Use sidekiq.yml while starting the sidekiq worker

sandip-mane:
1. The change passes the `sidekiq.yml` file to sidekiq process so that sidekiq will consider all configurations specified in `sidekiq.yml`.
2. Added unwanted files to `.slugignore` to ignore it for the slug on heroku.


sandip-mane:

PR #533, opened 2021-07-08
Removed slim and replaced the templates with erb

### LINKED ISSUE #532: Remove slim gems
We are using following two gems.

```
gem "slim"
gem "slim-rails"
```

We hardly have only a few files. It's not worth adding two gems for a few files. 

Secondly adding "embed" code with `script` tag is much more difficult in a slim file compared to an `.erb` file.

@akhil-gautam _a

akhil-gautam:
Fixes #532


akhil-gautam:

neerajdotname:
I verified it on heroku.

PR #545, opened 2021-07-21
Rubocop enforce space around operators

sandip-mane:
1. Rubocop, enforcing space around operators
2. Fixes for the following warning
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,16 +5,13 @@
   "lint-staged": {
     "app/**/*.{html,md,js,jsx,json}": [
       "prettier --write",
-      "eslint --fix",
-      "git add"
+      "eslint --fix"
     ],
     "{app,test}/**/*.rb": [
-      "bundle exec rubocop -a",
-      "git add"
+      "bundle exec rubocop -a"
     ],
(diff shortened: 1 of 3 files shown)


sandip-mane:

package.json
```diff
@@ -5,16 +5,13 @@
...
+      "eslint --fix"
     ],
     "{app,test}/**/*.rb": [
-      "bundle exec rubocop -a",
-      "git add"
+      "bundle exec rubocop -a"
     ],
     "Gemfile": [
-      "bundle exec rubocop -a",
-      "git add"
```
neerajdotname, line 17: @sandip-mane without the `git add` we need to manually add all the fixed files. I think it's a good idea to keep `git add`.

sandip-mane: @neerajdotname it works without `git add` too, husky takes care of it. I have verified it with a commit with changes in both `js` and `rb` files. 

I removed it because I saw the warning that's shown in the description. We can undo this change as it's just a warning.

neerajdotname: @sandip-mane ok sounds good.

PR #548, opened 2021-07-25
Added rubocop eslint and prettier hooks with modified configs

### LINKED ISSUE #547: Fix Rubocop, Eslint, Prettier config in order to port it to book
@Amaljith-K _a
Please take a look into this video: https://youtu.be/f4MCdrSW8jw

And first port the rubocop config, eslint, prettier config that we were planning to add in granite over here. Let's merge the PR here after taking review from Neeraj and then port it to granite.

yedhink:
Fixes #547 
- Also applied the new configs on existing files

@yedhink _a


.rubocop.yml
```diff
@@ -121,22 +292,22 @@ Layout/SpaceInsideBlockBraces:
...
-  Enabled: false
+  Enabled: true
   EnforcedStyle: double_quotes
 
-# Detect hard tabs, no hard tabs.
-Layout/IndentationStyle:
+# Requires a single final new line to the file
+Layout/TrailingEmptyLines:
   Enabled: true
+  EnforcedStyle: final_newline
```
neerajdotname, line 307: @yedhink I've seen rubocop complain that final new line is not there. If rubocop can automatically fix it then this is fine. If rubocop can't fix it then I find it unnecessary to manually add a blank line and in that case I would insist on disabling this cop.

.rubocop.yml
```diff
@@ -121,22 +292,22 @@ Layout/SpaceInsideBlockBraces:
...
 
-# This cop looks for trailing blank lines and a final newline in the source code.
-# This is being disabled because the requirement to have a final newline is illogical.
-Layout/TrailingEmptyLines:
-  Enabled: false
+Layout/EmptyLines:
+  Enabled: true
```
neerajdotname, line 310: Not sure what's the difference between this cop and the last cop but I guess I captured my concern in the last comment.

PR #559, opened 2021-08-04
Added changes to rubocop and eslint configuration

### LINKED ISSUE #563: Remove insert newline from vscode config

### LINKED ISSUE #566: Enforce leading dots while method chaining etc
Refer: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/DotPosition

sandip-mane:
[Edited by Yedhin] Fixes #566, Fixes #563

@yedhink Please check the updated configurations for rubocop and eslint.

**ESLINT**
`"react/no-unescaped-entities": 0` will not mandate unescaped characters in strings like `", >` etc.

**RUBOCOP**
Added
```
Layout/MultilineMethodCallIndentation:
#   Enabled: true
#   EnforcedStyle: indented_relative_to_receiver
```
Removed
```
Layout/TrailingEmptyLines:
  Enabled: true
  EnforcedStyle: final_blank_line
```
This was creating 2 blank lines at the end. Just having this check enabled will require a single line at the end.

Let me know your opinion.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -195,9 +195,9 @@ Layout/IndentationConsistency:
 #
 #   # do something
 # end
-Layout/MultilineMethodCallIndentation:
-  Enabled: true
-  EnforcedStyle: indented
+# Layout/MultilineMethodCallIndentation:
+#   Enabled: true
+#   EnforcedStyle: indented_relative_to_receiver
 
 # This cop ensures the indentation of the first parameter in a method definition.
 Layout/FirstParameterIndentation:
@@ -389,7 +389,6 @@ Layout/EmptyLineAfterGuardClause:
 # # EOF
 Layout/TrailingEmptyLines:
   Enabled: true
-  EnforcedStyle: final_blank_line
 
 # This cop checks for two or more consecutive blank lines.
 # This rule is not same as TrailingEmptyLines, because:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,10 +47,6 @@ AllCops:
 Layout/EmptyLinesAroundAccessModifier:
   Enabled: true
 
-Layout/DotPosition:
-  Enabled: true
-  EnforcedStyle: leading
-
 # This cop checks for redundant access modifiers, including those with no code,
 # those which are repeated, and leading `public` modifiers in a class or module body.
 Lint/UselessAccessModifier:
@@ -132,6 +128,12 @@ Layout/EmptyLinesAroundClassBody:
 # ====================================================================================================
 # All Method related rules
 # ====================================================================================================
+# This cop checks the . position in multi-line method calls.
+# The dot should be leading rather than trailing.
+Layout/DotPosition:
+  Enabled: true
+  EnforcedStyle: leading
+
 # No space in method name and the arguments
 Lint/ParenthesesAsGroupedExpression:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,6 +47,10 @@ AllCops:
 Layout/EmptyLinesAroundAccessModifier:
   Enabled: true
 
+Layout/DotPosition:
+  Enabled: true
+  EnforcedStyle: leading
+
 # This cop checks for redundant access modifiers, including those with no code,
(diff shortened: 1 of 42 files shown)


yedhink:
@yedhink _a

.rubocop.yml
```diff
@@ -195,9 +195,9 @@ Layout/IndentationConsistency:
 #
 #   # do something
 # end
-Layout/MultilineMethodCallIndentation:
-  Enabled: true
-  EnforcedStyle: indented
+# Layout/MultilineMethodCallIndentation:
+#   Enabled: true
+#   EnforcedStyle: indented_relative_to_receiver
```
yedhink, line 200: I didn't understand why this rule was removed. Can you point out what issues this rule was causing?

This is a good rule in my opinion.

The following is bad/unformatted code in my opinion:

```ruby
    def send_devise_notification(notification, *args)
      devise_mailer
      .send(notification, self, *args)
      .deliver_later(queue: "devise_email")
    end
```

This rule will correct and format it and make it more readable like so:
```ruby
    def send_devise_notification(notification, *args)
      devise_mailer
        .send(notification, self, *args)
        .deliver_later(queue: "devise_email")
    end
```

I suggest we keep this rule as it was.

.rubocop.yml
```diff
@@ -389,7 +389,6 @@ Layout/EmptyLineAfterGuardClause:
 # # EOF
 Layout/TrailingEmptyLines:
   Enabled: true
-  EnforcedStyle: final_blank_line
```
yedhink, line 392: Removing this is fine.

But I would like to confirm this. Do we need a blank line at end of files or do we end with code/characters or more specifically a `final_newline`?

Note that `newline` and `blank line` are two different entities. Refer to the examples in [this doc](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/TrailingEmptyLines).

yedhink (commented):
@sandip-mane _a

Let me know your thoughts on the following.

.rubocop.yml
sandip-mane: With the earlier code

**good**
```ruby
user_name = User.
  where(type: :bot).
  first.
  name
```

**bad**
```ruby
user_name = User
              .where(type: :bot)
              .first
              .name
```

IMO the second one is more readable. That's why added `indented_relative_to_receiver`.

.rubocop.yml
sandip-mane:
```yml
Layout/TrailingEmptyLines:
  Enabled: true
```
When we keeping only this will require one empty line after the code ends. If we add `final_blank_line`, it'll need one more blank line at the end. 

For reference, you can see the changes done in the `.rb` files.

.rubocop.yml
neerajdotname: If rubocop can autofix it then I'll go with whatever rubocop can do without caring about what the style it. That's because with rubocop atleast we have some consistency. Without rubocop some folks will do this way and some folks will do that way.

yedhink:
@sandip-mane 

Please watch the video and let me know your thoughts.

https://youtu.be/2Wgj7rXxDLg

One thing I missed to mention in the video is that trailing dots will also break syntax highlighting in VSCode/Text Editors in general. Thus we can enforce leading dots over trailing dots.

cc: @neerajdotname

neerajdotname:
@yedhink Thanks for the detailed video. I agree with what's presented there.

sandip-mane:
@yedhink 

I first changed it to `indented_relative_to_receiver` because I applied this rule in neetoDesk instead of trying out sample examples. In actual code, it looks bad.

**Bad**
image

**Good**
image

**Then I saw another problem,** (with `indented_relative_to_receiver` which I added)
When the variable name doesn't end at an even position, the rubocop auto-indents the methods on next line from 2 spaces after its receiver. Which is not indented properly.

In the screenshot, it's indenting it wrong, ideally, there should be either one more space or one less space.

Screenshot 2021-08-09 at 3 48 57 PM

-------------

But anyway, I have reverted the change. 
Let me know if you feel other changes are a problem too.

PS: I also think something's not set up correctly with your VSCode, I dont lose selection on giving a tab.

PR #575, opened 2021-08-12
Cleaned up sidekiq initializer and routes

### LINKED ISSUE #552: Fix sidekiq initializer
Shouldn’t the default value for `host` be `localhost` if no ENV variable is set? @sandip-mane

https://github.com/bigbinary/wheel/blob/f0a409c27b4f9282a5ef32ae45e319cfdfceffdf/config/initializers/sidekiq.rb#L10

https://github.com/bigbinary/wheel/blob/f0a409c27b4f9282a5ef32ae45e319cfdfceffdf/config/initializers/sidekiq.rb#L17

neerajdotname:
Fixes #552

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,20 +1,27 @@
 # frozen_string_literal: true
 
+Redis.exists_returns_integer = true
+
 if Rails.env.heroku?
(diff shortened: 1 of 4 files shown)


neerajdotname:
@sandip-mane _a Can you review it.

PR #583, opened 2021-08-16
Testing running rubocop for only changed files in semaphore

yedhink:
Please don't merge. Only for testing purposes.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -41,7 +41,9 @@ blocks:
 
         - name: Rubocop
           commands:
-            - bundle exec rubocop
+            - git log
+            - git branch -r --all --sort=-committerdate
+            - git diff-tree -r --no-commit-id --name-only HEAD master | xargs bundle exec rubocop --force-exclusion
 
   - name: Tests
     task:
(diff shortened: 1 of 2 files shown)


yedhink:
@neerajdotname @unnitallman 

Please go through the video and let me know your thoughts.

https://youtu.be/0JDmwU2lj0M

neerajdotname:
@yedhink @Vishal24367 //cc @unnitallman https://www.loom.com/share/575311efdfee465bac6be1439ecf436c

neerajdotname:
I thought this was related to AceInvoice somehow.

unnitallman:
@yedhink @neerajdotname 

1. I think that we should run rubocop on all the files in the CI.
2. It is safe to  run `rubocop -a` and send a single PR for that. if the test suite passes and with something manual tests, we can safely merge it. I don't foresee any issues coming out of this.
3. If only a small number of issues (5 as @yedhink says) remain, which needs to be fixed manually, I think its worth creating an issue and getting it fixed and then letting Rubocop run on all files everytime in the CI.

The work around (with a temp file) that Yedhin suggested is indeed hacky and not reliable. We may be able to find out better and reliable ways to make Rubocop run only on changed files - but I think we should just fix all the errors and then let Rubocop run on all the files.

neerajdotname:
@unnitallman sounds good.

PR #584, opened 2021-08-17
Fixes ESLint's issue with require

sandip-mane:
@yedhink was there any specific reason to turn off the following checks?

```js
    "no-unused-vars": "off",
    "no-undef": "off",
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,6 +36,8 @@ module.exports = {
     "no-console": "error",
     "import/prefer-default-export": "off",
     "react/prop-types": "off",
+    "no-unused-vars": "off",
+    "no-undef": "off",
     "react/no-unescaped-entities": "off"
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,8 +31,8 @@ module.exports = {
   parser: "babel-eslint",
   plugins: ["react", "prettier"],
   rules: {
-    indent: ["error", 2, { SwitchCase: 1 }],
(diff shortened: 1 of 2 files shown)


yedhink:
@sandip-mane Yes. In Cypress tests and all, we would have undefined variables. The rule was turned off because a lot of people in adding tests were facing issues. Espcially in Aceinoivce.

About the unused variables - in some cases, like say function definition, etc, we'd have props that are currently unused, but may be used in later progressions. It's better to let it stay rather than remove it. This is very relevant when passing is props to a component and someone else having to build on top of it.

If those unused vars need absolutely be removed then the reviewer can explicitly point it out. Example: A prop was removed from a component like neetoUi Button, then we'd have to ask the developer to remove passing in that prop.

I suggest we close this PR without merging. Please close this PR if you agree with me. Else let me know your thoughts. Thanks.

yedhink:
Also, I think we should remove the `env` field from package.json. I couldn't find any documentation regarding that field. Let me know if there is any relevance to that field in package.json.

sandip-mane:
Done. Reverted.

sandip-mane:
For ESLint to understand `require()` and `define()` when `no-undef` is turned on, we need either `node` or `amd` in ENV.

sandip-mane:
I also strongly feel we shouldn't have removed 
```js
    "no-unused-vars": "off",
    "no-undef": "off",
```

For cypress tests, we can have a cascading configuration that can be different that the configuration at the root.

PR #587, opened 2021-08-23
Added import based eslint rules and applied it

### LINKED ISSUE #561: Add rules to ensure consistent react code practices via eslint+prettier
- [x] Add a newline after import statements.
- [ ] Add a newline before and after writing `useEffect` hook.
- [ ] When naming states use consistent semantics where we prefix setter with `set{StateVariableName}` format.
- [x] Sort the import statements as we do in neeto products - 3rd party, local absolute imports, local relative imports.
- [x] Always return a  value rather than making it implicit in JS side. Refer [this rule](https://eslint.org/docs/rules/consistent-return#require-return-statements-to-either-always-or-never-specify-values-consistent-return).
- [x] Add linebreaks after guard clauses or function calls within a scope. We don't need to enforce this for `switch statements.Refer [this rule](https://eslint.org/docs/rules/padding-line-between-statements).
- [ ] Prevent common typos. Refer [this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-typos.md).
- [x] Sort prop types alphabetically. [Refer](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md).

yedhink:
Fixes #561


yedhink:
@edwinbbu @karthiknmenon 

Can you guys take a look into the this PR and let me know your perspective about the rules which have been added?

I personally tried to stick to the react best practices book from BB, and tried to enforce the standards.

During PR reviews, some things have been repeatedly pointed out like maintaining the order of imports, etc. But doing it again and again won't scale.

I have added relevant(available) rules that will take care of most of the issues.

Let me know your thoughts on this, and if you have any suggestions for the addition/removal of rules. You can refer to the below links to understand more about each of the added rules.

References for rules:
- https://github.com/yannickcr/eslint-plugin-react#list-of-supported-rules
- https://github.com/import-js/eslint-plugin-import/#rules


One caveat about the new eslint config:
- When we add a new alias(in say `config/webpack/custom.js`), then we'd also have to add it to the `pathGroups` key of `import/order` rule in `.eslintrc.js`. Eslint has no effective way to order webpack alias imports. This was the only way to make things work. If we had adhered to a standard where webpack alias imports are all prefixed with an `@` symbol, like say `import { func } from "@constants/utils"`, then we would've had a way to make eslint understand which is a webpack alias and which is an external dependency by matching regex pattern for all paths that start with `@` symbol. Here in our case, in `import { func } from "constants/utils"`, understanding whether `constants` is an external module or is a webpack alias from eslint's perspective is not trivial.

edwinbbu:
@yedhink _a Let's keep react imports at the top. In order to so you can refer the below code which I used in bigbinary-website repo.

```
"import/order": [
      "error",
      {
        groups: [
          "builtin",
          "external",
          "internal",
          "parent",
          "sibling",
          "index",
        ],
        pathGroups: [
          {
            pattern: "react+(|-native)",
            group: "external",
            position: "before",
          },
        ],
        pathGroupsExcludedImportTypes: ["react"],
        alphabetize: {
          order: "asc",
          caseInsensitive: true,
        },
      },
    ],"import/order": [
      "error",
      {
        groups: [
          "builtin",
          "external",
          "internal",
          "parent",
          "sibling",
          "index",
        ],
        pathGroups: [
          {
            pattern: "react+(|-native)",
            group: "external",
            position: "before",
          },
        ],
        pathGroupsExcludedImportTypes: ["react"],
        alphabetize: {
          order: "asc",
          caseInsensitive: true,
        },
      },
    ]
```

yedhink:
@edwinbbu Can you let me know what the pattern `pattern: "react+(|-native)",` is actually looking for? If there is an import like `import {func} from "react-native-css"` then should it be at the top? Or do we only need to make only `react` and `react-native` imports at the top?


More specifically is it a substring match that we are looking for or whole word match?

yedhink:
@edwinbbu It's only matching `react`. Thus I'm wondering why we need to capturing group.

image

Did you mean to move all imports which has the `react` keyword in it to the top of the file? I'm not sure why we should do that.

I thought we are only moving `import {React, ...} from "react"` to the top of the file because that import wouldn't be required in most cases once we update React to v18. So in order to make it easier to remove those imports, we are moving only the "react" import to top.

Let me know your thoughts and what exactly we need to move to top.

yedhink:
I'm merging this now. Currently only `react` and `react-native` imports, without any substring match will be moved to top. This is something that should be documented in react plus book.

edwinbbu:
Looks good. We don't have to match `react-native` since we won't be having the react native imports in this project.

PR #589, opened 2021-08-24
Updated Toastr message logic

### LINKED ISSUE #588: Something went wrong error message is displaying for invalid credentials
https://loom.com/share/42d1cf9d867d45199d0a186fa1544b8b

edwinbbu:
Fixes #588 

Screenshot :
Screenshot 2021-08-24 at 10 23 09 PM

Reason for issue:

Logic in neetoui
Screenshot 2021-08-24 at 10 23 54 PM

From the logic, you can see that the `Something went wrong.` is the fallback message.

We are not getting the errorObject as the instance of the `Error`. 
`Promise.reject(error)` code below makes it the instance of Error hence we have wrap inside the json like { error: message}

It is not breaking in other applications since we are showing Toastr in the individual places rather than in the inceptor.

@yedhink _a Please review

### DIFF
--- a/app/javascript/src/apis/axios.js
+++ b/app/javascript/src/apis/axios.js
@@ -38,9 +38,9 @@ const handleSuccessResponse = response => {
 const handleErrorResponse = (error, authDispatch) => {
   if (error.response?.status === 401) {
     authDispatch({ type: "LOGOUT" });
-    Toastr.error(error.response?.data?.error);
+    Toastr.error({ error: error.response?.data?.error });
   } else {
-    Toastr.error(error.response?.data?.error || error.message);
+    Toastr.error({ error: error.response?.data?.error || error.message });
   }
   return Promise.reject(error);
 };


yedhink:
@edwinbbu _a

I don't think passing in another `{error: "message"}` object as an argument to a chained method which is already named `Toastr.error` is a very intuitive pattern. People will have to dig through the neetoUI codebase in order to even figure this out in the first place. And also it adds a bit of redundancy too, given we are explicitly passing another made-up object just to satisfy the conditions.

This is something that should be taken care of in the neetoUI component itself.

Refer the following pseudo-code for `showErrorToastr`:
```js
  let errorMessage;
  if (isError(errorObject)) errorMessage = errorObject.message;
  else if (isAxiosErrorResponse(errorObject)) errorMessage = errorObject?.response?.data?.error;
  else if (errorObject.error) errorMessage = errorObject.error;
  else if (errorObject.errors) errorMessage = R.join("\n", errorObject.errors);
  else if (isNonEmptyString(errorObject)) errorMessage = errorObject;
  else errorMessage = "Something went wrong.";
```

This way we can invoke Toastr.error like so:
```js
const handleErrorResponse = (axiosError, authDispatch) => {
    Toastr.error(axiosError);
    return Promise.reject(axiosError);
};

or 

Toastr.error("My custom error message");
```

Also I didn't understand the real-world use case of these two conditions. I feel these two conditions shouldn't be present:

```js
  else if (errorObject.error) errorMessage = errorObject.error;
  else if (errorObject.errors) errorMessage = R.join("\n", errorObject.errors);
```

I mean why allow the developer to pass in argument in multiple ways? Wouldn't it create inconsistency? This will also create preferential invocation of the function, meaning some may pass in `{error:}` or some may pass in `{errors:}` etc.

With the above conditions people can end up calling this method is multiple ways, which I think should be constrained:

```js
Toastr.error({error: "message"});
Toastr.error({errors: ["msg1", "msg2"  ...]});
Toastr.error(errorObject);
```

I personally feel the following is much more under our control:

```js
Toastr.error(errorObject); // we allow either axiosError or JS error objects
Toastr.error("string message");
```

Let me know your thoughts on this. I strongly feel we should update this in neetoUI.

PS: I understand that removal of `{error:}` and `{errors:}` can't be done in one quick transition, cause it may be used in some neeto projects. What I would suggest is updating the `Toastr.error` calls in those projects first, and then removing these conditions from neetoUI.

edwinbbu:
@yedhink _a 

```
else if (errorObject.error) errorMessage = errorObject.error;
else if (errorObject.errors) errorMessage = R.join("\n", errorObject.errors);
```
This condition is provided because there are someplace where the error is sent as an array and in some places as a string from the backend.

There are also places where the error is sent as { message: errorMessage } as well { error: errorMessage } from the backend in multiple projects. If there is a uniform pattern, then we can reduce the logic lot more.

The changes you mentioned in the logic can be taken up in v2 Toastr component. 
We have frozen the changes to v1. Let's merge the PR and we can fix the issue properly once the v2 is incorporated in the wheel.

yedhink:
Sure thing.

There are some differences of opinion. I will point this out in neetoUI.

PR #595, opened 2021-08-30
Upgrade semaphore ruby, node and postgres version

### LINKED ISSUE #550: Update semaphore config to Ruby 3

datt:
Fixes #550.


datt:
cypress _t

yedhink:
@datt 

In general, if you're making a version update then please notify me before merging. The reason is that LRRB content directly uses some of the configs from wheel. Directly in the sense, we `curl` it.

Version updates, such `NodeJS` version update as done in this PR can have a wider impact because some devDependencies may change their versions and we'd have to ensure things work as we have mentioned in the book.

In the current PR, there are no such issues. But I'm saying this so that we can avoid such cases in the future.

The following are the key cases where I need to be notified before merging: `version updates of any kind`, `config updates - say eslint rubocop prettier`, `git hook updates`, `circleci config`, `tailwind config` etc.

cc: @sandip-mane

datt:
@yedhink I made the versions consistent with the chat repo. I was not aware of LRRB, will let you know next time.

PR #606, opened 2021-09-08
Added ignored patterns in line-length config

### LINKED ISSUE #605: Rubocop line-length rule flags false positives as offenses
https://github.com/bigbinary/wheel/blob/d79db6e88d527052812c13ff570843ada7b2525b/.rubocop.yml#L429-L431

This will flag long strings, long comments, and long regex values (that are intrinsically single lined) as offenses. They aren't auto-correctible either.

We need this additional `IgnoredPatterns` configuration:

```yml
Layout/LineLength:
  Enabled: true
  Max: 120
  IgnoredPatterns: [
    '^\s*#', # line that begins with comment
    '^\s*"', # line that begins with double quote (long string literal)
    "^\\s*'", # line that begins with single quote (long string literal)
    '"$', # line that ends with double quote (long string variable assignment)
    "'$", # line that ends with single quote (long string variable assignment)
    '\/$', # line that ends with slash (long regex variable assignment)
  ]
```

Self assigning @Amaljith-K _a

Amaljith-K:
This avoids flagging false positives (long strings, long regexes, and long comments) as offenses.
Fixes #605

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -429,6 +429,14 @@ Layout/CaseIndentation:
 Layout/LineLength:
   Enabled: true
   Max: 120
+  IgnoredPatterns: [
+    '^\s*#', # line that begins with comment
+    '^\s*"', # line that begins with double quote (long string literal)
+    "^\\s*'", # line that begins with single quote (long string literal)
+    '"$', # line that ends with double quote (long string variable assignment)
+    "'$", # line that ends with single quote (long string variable assignment)
+    '\/$', # line that ends with slash (long regex variable assignment)
+  ]
 
 # Indent using two spaces
 Layout/IndentationWidth:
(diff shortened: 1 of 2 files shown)


yedhink:
@Amaljith-K  _a

For all PR's in wheel that updates any configuration or gem/package versions, you need to tag me. The reason is that we have to update the same changes in the book too. In book, we directly curl wheel configs. So if anything needs porting in the book, I have to be notified about such a change.

Remove the Rubocop disabler comment from this file and check if your new rules are working or not:
https://github.com/bigbinary/wheel/blob/709470db413116b76ff0b1ed013921b6c8646caf/config/initializers/devise.rb#L103

Amaljith-K:
Yes @yedhink _a
It works.
image

yedhink:
@Amaljith-K _a

Then please do remove the comment `# rubocop:disable Layout/LineLength ` from this line:
https://github.com/bigbinary/wheel/blob/709470db413116b76ff0b1ed013921b6c8646caf/config/initializers/devise.rb#L103

and assign this PR back to me.

Amaljith-K:
@yedhink _a applied the requested change

PR #616, opened 2021-09-21
Allowing sanitised html for dangerouslySetInnerHTML

sandip-mane:
This is required as we mostly use ActionText and we render the content using `dangerouslySetInnerHTML`. 
With the current eslint restrictions, those instances are not valid anymore. 

This PR allows sanitised content under `dangerouslySetInnerHTML`.

### Example:

**Bad:**
```jsx
<div dangerouslySetInnerHTML={{ __html: content }} />
```

**Good:**
```jsx
const sanitizer = dompurify.sanitize;
<div dangerouslySetInnerHTML={{ __html: sanitizer(content) }} />
```

@unnitallman _a please review.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -72,12 +72,7 @@ module.exports = {
     "jam3/no-sanitizer-with-danger": [
       2,
       {
-        wrapperName: [
-          "dompurify",
-          "sanitizer",
-          "sanitize",
-          "dompurify.sanitize"
-        ]
+        wrapperName: ["dompurify", "sanitizer", "sanitize"]
       }
     ],
     // not-auto-fixable: Report when a DOM element is using both children and dangerouslySetInnerHTML.
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,7 +37,7 @@ module.exports = {
     sourceType: "module"
   },
   parser: "babel-eslint",
-  plugins: ["react", "prettier", "import", "react-hooks", "promise"],
+  plugins: ["react", "prettier", "import", "react-hooks", "promise", "jam3"],
   rules: {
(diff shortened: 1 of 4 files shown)


yedhink:
@sandip-mane Once again kindly request you to please inform me if the configs in the wheel are updated.

I had already pointed out the reason why to inform me: https://github.com/bigbinary/wheel/pull/595#issuecomment-908887163

Currently, if any students, intern, an outsider who uses the Learn Ruby on Rails book tries out the configs specified there, say eslint, then it will fail for them because we haven't mentioned installing the new package that was added in this PR. It reflects really bad upon BigBinary itself given that we say that `everything in the book should work out of the box by just copy-pasting itself`.

@unnitallman 
Would you be able to add me as a code owner for `wheel`? I think that is the only way to overcome this issue.

sandip-mane:
@yedhink turning on notifications will help?

yedhink:
@sandip-mane Notifications are already turned on. I followed the BigBinary handbook and used the [control mechanism suggested](https://www.bigbinary.com/how-we-work/control-notifications-from-github#turn-off-unnecessary-github-notifications). Let me know if what's mentioned in the handbook is outdated and if it shouldn't be followed.

I can turn on notifications for all things that happen in Wheel. But still, it would defeat the purpose or the true issue that we are trying to counter here - that is LRRB breaking for readers when a breaking change is added to wheel.
Example:
- I sign-off at 6 PM.
- Another user adds a breaking change to wheel at 11pm at night.
- The notifications will be sent to my email. But LRRB will be down until I start working again next day.

Hope the gist is clear now.

The only solution here is to add me as a codeowner.

unnitallman:
@yedhink https://github.com/bigbinary/wheel/pull/618

PR #619, opened 2021-09-22
Modularized eslint configs and added ability for customizing certain eslint rules per project basis

### LINKED ISSUE #615: Add custom eslint option
@yedhink _a

### LINKED ISSUE #620: Logical expressions shouldn't be checked for indentation
@yedhink _a

yedhink:
Fixes #615, Fixes #620

@unnitallman

**Please don't merge yet**.

We should only merge this PR after the `sync_with_wheel` script is ready and neetoCompliance is updated.


yedhink:
@unnitallman 

The script is ready: https://gist.github.com/yedhink/d0e88d99536eca2bffb7becab49a0a39

Please note that this script will work once this PR gets merged. The reason is that the `.eslint-rules` folder gets added to master only once this PR is merged.

In neetoCompliance, the array of configs declared from the script that I've created can be used for auditing.

One more thing to notice here is that when we update the gist, the raw URL only gets updated after like 10 seconds due to caching issue in Github's side. We can counter it to an extent by passing an URL param with a random cache value. Thought I'd point this out.

We can merge this PR once changes are merged in neetoCompliance. Either ways neeto users will have to update neetoCompliance to latest version in order for changes to get reflected. Thus this PR can wait.

yedhink:
@unnitallman Also, [this is how](https://github.com/bigbinary/wheel/pull/619/commits/a38105ccbfea5c62bce6bdc74f12fe07cdb0259f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R9) we should add the `sync_with_wheel` command in `package.json`.

Users can run the following from terminal after adding above script to package.json to pull latest configs:

```bash
npm run sync_with_wheel
```

yedhink:
@unnitallman As per our discussion, I have moved the script into wheel itself.

This will allow us to update the script with much more ease when say a new config or package is added to wheel.

Will wait till neetoCompliance is ready to merge this PR.

unnitallman (approved):

PR #662, opened 2021-10-07
Added helpers to handle non-stringified values when parsing keys from localStorage

### LINKED ISSUE #661: Handle non stringified auth tokens which are not JSON parseable
Currently when setting `authToken` and `authEmail` to localStorage, the values are stringified. When reading these values from localStorage, we are expecting a stringified value. 

If non stringified tokens are already set in localStorage, we get an error trying to `JSON.parse()` these values.

Screen Shot 2021-10-07 at 11 07 11 AM

In order to fix this, when the value being read is not JSON parseable, we need to clear the specific tokens from localStorage instead.

CC: @yedhink 

@sreeram-venkitesh _a

sreeram-venkitesh:
Fixes #661 

Changes made
* Added localStorage helper at `javascript/src/helpers/storage.js`, similar to what is done in `granite`
* Added check inside `getFromLocalStorage` method to check if existing value is JSON stringified. If not, the value is replaced with `null`.
* Replaced all existing `localStorage.setItem()` and `localStorage.getItem()` with `setToLocalStorage()` and `getFromLocalStorage()` methods from the helper.

@yedhink _a

### DIFF
--- a/app/javascript/src/helpers/storage.js
+++ b/app/javascript/src/helpers/storage.js
@@ -7,8 +7,8 @@ const getFromLocalStorage = key => {
   try {
     storedValue = JSON.parse(localStorage.getItem(key));
   } catch (error) {
-    logger.error(error);
     localStorage.setItem(key, JSON.stringify(null));
+    logger.error(error);
   }
   return storedValue;
 };


yedhink:
@sreeram-venkitesh Also, the title of the PR needs to be in the past tense. We don't follow imperative present tense in commit messages nor PR titles.

I always give this reference to someone who starts new and is working with me: https://github.com/bigbinary/bigbinary-website/pull/781#pullrequestreview-677312016

PS: If we need to discuss or bring something to my attention or need my help, then the review comment can be left unresolved.

sreeram-venkitesh:
@yedhink _a

Changed the order of the return statement as mentioned in the review, in my last commit.

I also tried the steps you mentioned and tried switching from granite to wheel. Once I started wheel's server, reloading the logged in granite page gave the home screen for wheel. This is the intended behaviour if we already have `authToken` and `authEmail` stored from a different project. 

Please review the changes. Thank you.

PR #706, opened 2021-11-03
Added new eslint rule for hooks: react-hooks/exhaustive-dep

### LINKED ISSUE #705: Add new eslint rule for hooks: react-hooks/exhaustive-deps
With reference to the suggestion:
https://github.com/bigbinary/neeto-invoice-rn/pull/100#issuecomment-958312831

Other References:
https://reactjs.org/docs/hooks-rules.html
https://stackoverflow.com/questions/58866796/understanding-the-react-hooks-exhaustive-deps-lint-rule

neetoInvoice:
https://github.com/bigbinary/neeto-invoice-rn/pull/100

sangameshsomawar:
[Edited by Yedhin]
Fixes #705

### DIFF
--- a/.eslint-rules/react.js
+++ b/.eslint-rules/react.js
@@ -27,6 +27,6 @@ module.exports = {
     // not-auto-fixable: Ensures https://reactjs.org/docs/hooks-rules.html.
     "react-hooks/rules-of-hooks": "error",
     // not-auto-fixable: Ensures https://reactjs.org/docs/hooks-rules.html - Checks effect dependencies.
-    "react-hooks/exhaustive-deps": "off"
-  }
+    "react-hooks/exhaustive-deps": "error",
+  },
 };
--- a/app/javascript/src/components/Main.jsx
+++ b/app/javascript/src/components/Main.jsx
@@ -34,7 +34,7 @@ const Main = props => {
     initializeLogger();
     registerIntercepts(authDispatch);
     setAuthHeaders(setLoading);
-  }, []);
+  }, [authDispatch, props?.user, userDispatch]);
 
   if (loading) {
     return (
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-promise": "^5.1.0",
     "eslint-plugin-react": "^7.26.1",
-    "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint-plugin-react-hooks": "@next",
     "husky": "^7.0.2",
     "lint-staged": "^11.2.3",
     "postcss": "^8.2.15",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3742,10 +3742,10 @@ eslint-plugin-promise@^5.1.0:
   resolved "https://registry.yarnpkg.com/eslint-plugin-promise/-/eslint-plugin-promise-5.1.0.tgz#fb2188fb734e4557993733b41aa1a688f46c6f24"
   integrity sha512-NGmI6BH5L12pl7ScQHbg7tvtk4wPxxj8yPHH47NvSmMtFneC077PSeY3huFj06ZWZvtbfxSPt3RuOQD5XcR4ng==
 
-eslint-plugin-react-hooks@^4.2.0:
-  version "4.2.0"
-  resolved "https://registry.yarnpkg.com/eslint-plugin-react-hooks/-/eslint-plugin-react-hooks-4.2.0.tgz#8c229c268d468956334c943bb45fc860280f5556"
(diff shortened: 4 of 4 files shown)


yedhink:
@sangameshsomawar I had removed this ESLint rule on purpose because it was giving false positives in our neeto repos.

If you can make a strong case for including this rule and possible fixes for the issues that this rule points out in neeto repos, then I will consider adding it here.

PS: Moving forward, raise an issue and mention me in it, if you want to add something to wheel. It's better to first discuss the feasibility of adding something new before making a PR. Also please read: https://www.bigbinary.com/how-we-work/mention-closes-fixes-resolves

Closing this PR for now. You can raise a new issue in wheel, if you have a strong case to make.

sangameshsomawar:
@yedhink 

I would like to understand the false positives as I have never faced any.
We can meet over a call or you can simulate then share with me.

Regarding auto fix, **Eslint doesn't auto-fix if something changes the behavior of code.**
In our case, a list of exhaustive-deps really changes the behavior of code. So the developer has to decide it, eslint can't help. **But only because it doesn't auto-fix we should not overlook it.**


I highly agree that this rule looks tough in the beginning but it has its importance.

Thanks for pointing out the processes. I will take care of it next time 😄

yedhink:
@sangameshsomawar I saw issue #705 now and read through the references. Didn't know you had raised the issue since it wasn't linked in this PR.

What I had meant by false positives is that Main.jsx etc are receiving props that are given from the server itself rather than computed on the client-side. Thus if there is a change in those server props, then a re-render from scratch would be either way be required from my understanding.

But reading more about this rule and seeing that it's a good practice, I think we can enforce it and see what issues devs are facing based on this rule.

Reopening this PR.

.eslintrc.js
```diff
@@ -26,24 +26,25 @@ module.exports = {
...
     },
     ecmaVersion: 2018,
-    sourceType: "module"
+    sourceType: "module",
   },
   // babel-eslint is deprecated now. This is the latest package.
   parser: "@babel/eslint-parser",
   plugins: ["react", "prettier", "import", "react-hooks", "promise", "jam3"],
   rules: {
+    "react-hooks/exhaustive-deps": "error",
```
yedhink, line 47: We already have this rule `.esint-rules/react.js`. Make this fix over there.

yedhink (commented):
@sangameshsomawar _a

Once you add the rule, make sure to run it on all files within this repo.

Refer: https://www.bigbinary.com/learn-rubyonrails-book/linting-and-formatting-code#formatting-all-javascript-files

For the time being, I have updated the first comment in this PR to mention which issue it fixes. Moving forward please take care of it.

Please go through the comment and make the necessary changes.

.eslintrc.js
sangameshsomawar: Updated.

yedhink:
@sangameshsomawar _a

Few things to be followed:
- In BigBinary whenever we are making a PR, we assign the PR to the tech lead of that particular project.
- Once the TL reviews the PR, and if they request any changes, then they will assign the PR back to you
- And once you make the changes and push it, then you can assign the PR back to the reviewer.
- No need to mention the issue number in the commit message. We follow the `Squash and Merge` strategy. Also, we explicitly link the issue in the first comment of a PR.

Currently:
- [ ] This PR has conflicts with master. Thus rebase and fix the conflicts and push the changes.
- [ ] Go through the comments in this PR: https://github.com/bigbinary/wheel/pull/662#issuecomment-937575736
- [ ] I had explicitly pointed out something to be done in https://github.com/bigbinary/wheel/pull/706#pullrequestreview-796241998. Please carefully go through that comment and update with necessary changes.

Once all things are done, you can assign this PR back to me.

sangameshsomawar:
@yedhink 

Sorry if I am troubling you a lot. I am new here in BB 😄
Thank you, For guiding me through the process.

One more issue I am facing when I executed `npx eslint --fix "./app/javascript/src/**/*.{js,jsx,json}"`
```
Error: Rules with suggestions must set the `meta.hasSuggestions` property to `true`.
Occurred while linting /Users/sangameshsomawar/Desktop/bbprojects/wheel/app/javascript/src/components/Main.jsx:32
Rule: "react-hooks/exhaustive-deps"
```

**Root Cause Analysis:**
2 weeks ago, Someone upgraded eslint to v8.0.1 :  https://github.com/bigbinary/wheel/blob/master/package.json#L34

Eslint v8 has some breaking changes: 
https://eslint.org/docs/8.0.0/user-guide/migrating-to-8.0.0#suggestions

The above issues have been already reported & fixed by react team.
https://github.com/facebook/react/issues/22545

We need to wait for the new version of `eslint-plugin-react-hooks`.
We can use `eslint-plugin-react-hooks@next` but I don't think that will be safe.

yedhink:
@sangameshsomawar 

>Sorry if I am troubling you a lot. I am new here in BB

It's my pleasure to help you out. Feel free to ask any questions you have.

>We can use eslint-plugin-react-hooks@next but I don't think that will be safe.

This is the only option that I see. We can revert back to stable version once they publish the latest stable version.

yedhink (approved):

PR #744, opened 2021-12-06
Upgrade gcc version for semaphoreCI

chiraggshah:
By default, gcc v4 caused random failures for `bundle install` step.
Upgrading gcc to v8 fixes the issue.


chiraggshah:
@yedhink _a : Please merge this.

PR #755, opened 2022-01-04
Publish junit xml to semaphoreci

chiraggshah:


chiraggshah:
@yedhink _a : Minor changes to upload JUnit XML for tests to semaphoreCI

.semaphore/semaphore.yml
```diff
@@ -51,7 +51,9 @@ blocks:
 
         - name: Eslint
           commands:
-            - curl -s -L "https://raw.githubusercontent.com/bigbinary/wheel/master/.semaphore/commands/run_eslint_on_modified_files.sh" | bash
+            - curl -s -L
+              "https://raw.githubusercontent.com/bigbinary/wheel/master/.semaphore/commands/run_eslint_on_modified_files.sh"
+              | bash
```
yedhink, line 56: Is there a need to split this into newlines? I would advice against splitting a single command into newlines. The reason is that neetoCompliance might be most probably checking for regex match of a single line rather than multiline.

Most probably these newlines might have got added by some VSCode extension from local. In Wheel's git hooks we haven't added anything to format `yaml` files as far as I know.

Thus we should revert this change.

yedhink (commented):
@chiraggshah _a

I have added one minor comment. Please take a look. Thanks.

.semaphore/semaphore.yml
chiraggshah: Yup, looks like VSCode added those changes based on my local preferences. Will revert that part.

chiraggshah:
@yedhink _a : Done. Please review.

PR #761, opened 2022-01-13
Updated React code to react best practices

### LINKED ISSUE #758: Update React code with our React best practices
https://github.com/bigbinary/bigbinary-website/issues/1765

- Start by changing `Yup` to `yup`!
- Form validations should be written within `constants.js` in the closest scope

@rishabhanand12 _a

`wheel` is being used as a base for neetoUI challenge. People are following the same wrong conventions that is used in wheel.

Can you take this up with some priority? Thanks.

rishabhanand12:
Fixes #758 

@rishabhanand12 _a


rishabhanand12:
@yedhink _a Please take a look.

yedhink (changes requested):
@rishabhanand12 _a

Please watch: https://www.loom.com/share/babde97c593040008db28b51d456505a

rishabhanand12:
@yedhink _a Please review.

app/javascript/src/components/Authentication/Login.jsx
```diff
@@ -44,60 +48,56 @@ const Login = ({ history }) => {
...
+            <Button
+              fullWidth
+              type="submit"
+              label="Login"
+              data-cy="login-submit-button"
+              className="h-8"
+              loading={loading}
+              onClick={() => {
+                setSubmitted(true);
+              }}
```
yedhink, line 83:
```suggestion
              onClick={() => setSubmitted(true)}
```

app/javascript/src/components/Authentication/Login.jsx
```diff
@@ -44,60 +48,56 @@ const Login = ({ history }) => {
...
+            />
+          </Form>
         </Formik>
         <div className="flex flex-col items-center justify-center mt-4 space-y-2">
           <div className="flex flex-row items-center justify-start space-x-1">
             <p className="font-normal text-gray-600">{`Don't have an account?`}</p>
```
yedhink, line 89: Please remove the template strings. It's an anti-pattern when using it like this in JSX.

app/javascript/src/components/Authentication/ResetPassword.jsx
```diff
@@ -25,39 +25,35 @@ const ResetPassword = () => {
...
+            id="new_user"
+          >
+            <Input name="email" label="Email" type="email" required />
+            <div className="flex flex-col items-center justify-center space-y-2">
+              <Button
+                fullWidth
+                type="submit"
+                onClick={() => {
+                  setSubmitted(true);
+                }}
```
yedhink, line 45: Refer prev comments.

yedhink (changes requested):
@rishabhanand12 _a

I have added some comments. Please go through them and let me know if you need any clarifications.

rishabhanand12:
@yedhink _a Please take a look.

app/javascript/src/components/Authentication/Signup.jsx
```diff
@@ -43,67 +35,63 @@ const Signup = ({ history }) => {
...
+              type="password"
+              label="Confirm password"
+              placeholder="******"
+            />
+            <Button
+              fullWidth
+              type="submit"
+              onClick={() => {
+                setSubmitted(true);
+              }}
```
yedhink, line 85: Not sure why you haven't made this change although pointed out in review comments.
```suggestion
              onClick={() => setSubmitted(true)}
```

app/javascript/src/components/Dashboard/Account/Passwords/Edit.jsx
```diff
@@ -43,32 +33,38 @@ const Edit = () => {
...
+              fullWidth
               name="submit"
               type="submit"
               label="Update"
               className="h-8"
-              fullWidth
+              loading={loading}
+              onClick={() => {
+                setSubmitted(true);
+              }}
```
yedhink, line 67: Same as prev comment.

app/javascript/src/components/Dashboard/Account/Profile.jsx
```diff
@@ -0,0 +1,54 @@
...
+              name="password"
+              label="Current password"
+              type="password"
+            />
+            <Button
+              fullWidth
+              type="submit"
+              onClick={() => {
+                setSubmitted(true);
+              }}
```
yedhink, line 43: Same

app/javascript/src/components/Hero.jsx
```diff
@@ -2,21 +2,21 @@ import React from "react";
...
-    "Uses slim for cleaner syntax over erb and better performance over haml.",
-    "Intercepts all outgoing emails in non production environment using gem mail_interceptor.",
-    "Uses SemaphoreCI for continuous testing.",
-    "Content compression via Rack::Deflater.",
-    "Auto-formats Ruby code with rubocop.",
-    "Auto-formats JavaScript and CSS code with prettier.",
-    "Performs background job processing 'inline' for heroku env. It means heroku can deliver emails.",
-    "Letter opener gem for development.",
-  ];
+import { FEATURES_LIST } from "./constants";
```
yedhink, line 5: LIST is already plural:
```suggestion
import { FEATURE_LIST } from "./constants";
```

yedhink (changes requested):
@rishabhanand12 _a

Few comments added. Carefully go through them and make the changes wherever necessary within this PR.

rishabhanand12:
@yedhink _a I have made the suggested changes. Please review.

app/javascript/src/components/Authentication/Signup.jsx
```diff
@@ -41,53 +38,54 @@ const Signup = ({ history }) => {
...
+                type="email"
+                label="Email"
+                placeholder="oliver@example.com"
+              />
+              <Input
+                required
+                name="firstName"
+                type="text"
+                label="First name"
+                placeholder="Sam"
```
yedhink, line 55:
```suggestion
placeholder="Oliver"
```

yedhink (changes requested):
@rishabhanand12 _a

One minor fix needed.

yedhink:
@rishabhanand12 _a

https://github.com/bigbinary/wheel/pull/763#issuecomment-1016537379

rishabhanand12:
@rishabhanand12 Add a new tab for updating email in the update profile tab. Users should be logged out on updating `email` and `password`. 

cc: @yedhink

rishabhanand12:
@yedhink If we are using tabs inside the profile menu for profile update and email update then should the page URL reflect the selected tab?

yedhink:
@rishabhanand12 _a

Yes. We should have some sort of route identification for each tab. Query string would do.

This will help us in navigating to specific tab if need be.,

yedhink (changes requested):
@rishabhanand12 _a

https://www.loom.com/share/aab69e01e7bf4633a773321e9e3f9a99

yedhink:
@rishabhanand12 
Also, we had already made changes related to using bang methods in granite: https://github.com/bigbinary/granite/blob/b3d866eb14d042e132ce77a6672029be0d10aa92/app/controllers/application_controller.rb#L4-L7

Follow the same pattern here too.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -5,13 +5,12 @@ class Api::V1::SessionsController < Api::V1::BaseController
...
-    user = User.find_for_database_authentication(email: params[:user] && params[:user][:email])
+    user = User.find_for_database_authentication(email: session_params[:email])
     if invalid_password?(user)
-      respond_with_error "Incorrect email or password", 401
+      respond_with_error(t("invalid_credentials"), :unauthorized)
     else
       sign_in(user)
-      render json: { auth_token: user.authentication_token, user: user, is_admin: user.super_admin? },
-        location: root_path, status: :created
+      render status: :created, json: { auth_token: user.authentication_token, user: user, is_admin: user.super_admin? }
```
yedhink, line 13: Use respond_with_success helper that we created.

app/controllers/api/v1/users_controller.rb
```diff
@@ -1,57 +1,28 @@
 # frozen_string_literal: true
 
 class Api::V1::UsersController < Api::V1::BaseController
-  skip_before_action :authenticate_user!, only: [:create]
-  skip_before_action :authenticate_user_using_x_auth_token, only: [:create]
+  skip_before_action :authenticate_user!, only: :create
+  skip_before_action :authenticate_user_using_x_auth_token, only: :create
 
-  before_action :set_user, only: [:show, :edit, :update, :destroy]
+  before_action :set_user!, only: %i[show destroy]
```
yedhink, line 7: Use `load` keyword. Make changes wherever necessary.

app/controllers/api/v1/users_controller.rb
```diff
@@ -1,57 +1,28 @@
...
 
   def show
-    if @user
-      render json: @user
-    else
-      respond_with_error "User with id #{params[:id]} not found.", :not_found
-    end
+    render status: :ok, json: @user
```
yedhink, line 10: Use respond_with_success. Make changes wherever necessary.

app/controllers/concerns/api_rescuable.rb
```diff
@@ -0,0 +1,40 @@
...
+    def handle_api_error(exception)
+      log_exception(exception)
+      respond_with_error(exception.record.errors_to_sentence, :internal_server_error)
+    end
+
+    def log_exception(exception)
+      Rails.logger.info exception.class.to_s
+      Rails.logger.info exception.to_s
+      Rails.logger.info exception.backtrace.join("\n")
+    end
```
yedhink, line 39: Move this to a `Loggable` concern.

app/controllers/concerns/authenticable.rb
```diff
@@ -0,0 +1,24 @@
...
+      user_email = request.headers["X-Auth-Email"]
+      auth_token = request.headers["X-Auth-Token"].presence
+      user = user_email && User.find_by_email(user_email)
+
+      if user && Devise.secure_compare(user.authentication_token, auth_token)
+        sign_in user, store: false
+      else
+        respond_with_error("Could not authenticate with the provided credentials", 401)
+      end
+      end
```
yedhink, line 23: Fix formatting.

app/controllers/concerns/authenticable.rb
```diff
@@ -0,0 +1,24 @@
...
+    before_action :authenticate_user_using_x_auth_token
+    before_action :authenticate_user!
+  end
+
+  private
+
+    def authenticate_user_using_x_auth_token
+      user_email = request.headers["X-Auth-Email"]
+      auth_token = request.headers["X-Auth-Token"].presence
+      user = user_email && User.find_by_email(user_email)
```
yedhink, line 16: Don't use dynamic finder methods. Make changes wherever necessary.

https://www.bigbinary.com/learn-rubyonrails-book/find-vs-find-by-vs-where-in-active-record#don-t-use-dynamic-finder-methods

app/controllers/registrations_controller.rb
```diff
@@ -1,33 +1,25 @@
 # frozen_string_literal: true
 
 class RegistrationsController < Devise::RegistrationsController
-  prepend_before_action :authenticate_scope!, only: [:update_password]
-  before_action :load_resource, only: [:update_password]
+  prepend_before_action :authenticate_scope!, only: [:update]
+  before_action :load_resource, only: [:update]
```
yedhink, line 5: Use %i. Make it a practice.

app/controllers/registrations_controller.rb
```diff
@@ -1,33 +1,25 @@
...
 
-  def update_password
-    if resource.update_with_password(password_update_params)
+  def update
+    if resource.update_with_password(account_update_params)
       bypass_sign_in resource, scope: :user
-      render json: { notice: "Password has been successfully updated" }, status: :ok
+      render status: :ok, json: { notice: "User profile has been successfully updated" }
```
yedhink, line 10: Use api response helper and also move this string to en.yml

app/javascript/src/components/Authentication/Signup.jsx
```diff
@@ -43,67 +32,64 @@ const Signup = ({ history }) => {
...
               <Input
+                required
                 name="password"
                 type="password"
                 label="Password"
                 placeholder="******"
-                required
               />
               <Input
+                required
```
yedhink, line 72: Do we need this `required` statement over here if the Yup validation already requires it? Can you check on how it's different?

app/javascript/src/components/Dashboard/Notes/Table.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from "react";
+
+import { Table as NeetoUITable } from "neetoui";
+
+import { NOTES_TABLE_COLUMN_DATA } from "./constants";
+import EditNotePane from "./Pane/Edit";
+
+export default function Table({ setSelectedNoteIds, notes = [], fetchNotes }) {
```
yedhink, line 8: Please don't use `function`. Make this an arrow function.

app/javascript/src/components/Hero.jsx
```diff
@@ -0,0 +1,37 @@
...
+        <li key={index} className="py-3 border-b border-gray-100 last:border-0">
+          {feature}
+        </li>
+      ))}
+    </ul>
+  );
+};
+
+const Hero = () => {
+  return (
```
yedhink, line 20: Make this implicit return.

app/models/user.rb
```diff
@@ -11,7 +11,7 @@ class User < ApplicationRecord
   before_save :ensure_authentication_token_is_present
 
   validates :first_name, :last_name, :email, presence: true
-  validates :email, uniqueness: true
+  validates_uniqueness_of :email, case_sensitive: false
```
yedhink, line 14: Revert this change. Pass in `case_sensitive: false` to uniqueness key hash.

yedhink (changes requested):
@rishabhanand12 _a

Few comments that you can work on before assigning me back.

app/controllers/passwords_controller.rb
```diff
@@ -6,7 +6,7 @@ class PasswordsController < Devise::RegistrationsController
 
   def update_password
     if resource.update_with_password(password_update_params)
-      bypass_sign_in resource, scope: :user
+      sign_out(resource)
```
yedhink, line 9: Is this the forgot password scenario? If it's changing the password from profile scenario, then we shouldn't sign out the user when password is updated. Bypass the sign in.

Ignore if both of us are on the same page.

Read: https://security.stackexchange.com/a/178669

app/controllers/profiles_controller.rb
```diff
@@ -14,9 +14,19 @@ def update
     end
   end
 
+  def update_email
+    if resource.update_with_password(update_params)
+      sign_out(resource)
```
yedhink, line 19: Here it's unavoidable I guess because of our localStorage also need to be updated with latest email etc and so the contexts. So before updating the email, in the UI make sure to show an alert that this will sign them out.

app/controllers/profiles_controller.rb
```diff
@@ -14,9 +14,19 @@ def update
...
+      render status: :ok, json: { notice: "Email has been successfully updated" }
+    else
+      clean_up_passwords resource
```
yedhink, line 22: Not sure why we need to do this here.

app/javascript/src/components/Dashboard/Settings/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const [activeNavlink, setActiveNavlink] = useState(
+    getActiveNavLink(tab) || SETTINGS_NAVLINKS[0]
+  );
+
+  useEffect(() => history.push(activeNavlink?.path), [activeNavlink]);
+
+  return (
+    <>
+      <MenuBar showMenu={true} title="Settings">
```
yedhink, line 20:
```suggestion
<MenuBar showMenu title="Settings">
```

config/routes.rb
```diff
@@ -5,12 +5,13 @@ def draw(routes_name)
...
 
-  devise_for :users, path_prefix: "devise", controllers: { registrations: "registrations" }
+  devise_for :users, path_prefix: "devise", controllers: { registrations: "profiles" }
 
   devise_scope :user do
     scope "my" do
-      put "profile/update", to: "registrations#update"
-      put "password/update", to: "passwords#update_password"
+      put "profile", to: "profiles#update"
+      patch "password", to: "passwords#update_password"
```
yedhink, line 13: Why are we calling this action `update_password`? Is `update` action of `passwords` controller used for some other purpose?

yedhink (changes requested):
@rishabhanand12 _a

Few comments.

rishabhanand12:
@yedhink I have verified that the tests are passing locally. Could you please give me access to semaphore logs.

app/controllers/passwords_controller.rb
yedhink: Then we should not sign out the user.

yedhink:
@rishabhanand12 _a Let's try to merge this PR today. Lemme know if you have any blockers.

app/controllers/passwords_controller.rb
rishabhanand12: @yedhink Should we use a field which the user can select to log out of all active sessions? 

Also, do you want me to make similar changes to the email form?

yedhink: @rishabhanand12 

>Should we use a field which the user can select to log out of all active sessions?

Let's take that up in another issue. Not in this one. Because we'd have to regenerate the auth token for the user for logging out of all devices.

>Also, do you want me to make similar changes to the email form?

No. Email form, we have to log the user out.

rishabhanand12:
@yedhink _a Please review.

app/controllers/profiles_controller.rb
```diff
@@ -1,26 +1,26 @@
...
+  include ApiResponders
+
   prepend_before_action :authenticate_scope!, only: :update
   before_action :load_resource
 
   def update
     if resource.update_with_password(update_params)
       bypass_sign_in resource, scope: :user
-      render status: :ok, json: { notice: "User profile has been successfully updated", user: resource }
+      respond_with_success(t("successfully_updated", entity: "User profile"), :ok, { user: resource })
```
yedhink, line 12: No need to pass in `ok`. That should be the default value.

app/javascript/src/components/Dashboard/Settings/Email.jsx
```diff
@@ -50,12 +62,21 @@ const Email = () => {
...
-                onClick={() => setSubmitted(true)}
+                onClick={e => showUpdateConfirmation(e, validateForm)}
                 label="Update"
                 className="h-8"
                 loading={isSubmitting}
                 disabled={isSubmitting}
               />
+              <Alert
+                isOpen={showUpdateAlert}
+                message="Are you sure you want to continue? You will need to login to continue."
```
yedhink, line 73:
```suggestion
                message="Are you sure you want to update your email?"
```

test/controllers/passwords_controller_test.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class PasswordsControllerTest < ActionDispatch::IntegrationTest
+  def setup
+    @nancy = create(:user, :nancy)
+    sign_in @nancy
```
yedhink, line 8: Use Oliver and Eve as personas.

yedhink (changes requested):
@rishabhanand12 _a

PR looks great. Few comments. No need to rush on these changes. Let's merge this by Monday.

app/controllers/profiles_controller.rb
rishabhanand12: @yedhink That would work if `status` was the last argument. If we don't pass `:ok`, then the user json that we pass will be considered to be the status which will lead to an error.

rishabhanand12:
@yedhink _a Please take a look. I have made a couple of more changes to the user and notes factories based on PR reviews.

yedhink:
@rishabhanand12 _a

I am getting an error while running `./bin/setup` for this branch:
image


Take a look.

yedhink:
@rishabhanand12 _a

Once we signup, there needs to be Toastr showing whether signup was successful or not.

yedhink:
@rishabhanand12 _a

https://www.loom.com/share/8ca87653333d4f47925d6538fdf97e43

yedhink:
@rishabhanand12 _a

Remove unnecessary pages. Use existing tabs:
https://www.loom.com/share/8e51e63597234e308a7d26fcee86bd7c

app/javascript/src/components/Dashboard/Settings/Profile.jsx
```diff
@@ -9,25 +9,37 @@ import profilesApi from "apis/profiles";
...
+  const handleSubmit = async (data, { resetForm }) => {
     try {
       const {
         data: { user },
       } = await profilesApi.updateProfile(data);
       userDispatch({ type: "SET_USER", payload: { user } });
     } catch (err) {
       logger.error(err);
+    } finally {
+      resetForm("password");
```
yedhink, line 31: If we are resetting form over, then I think we can add enableReinitialize to the formik form.

- https://formik.org/docs/api/formik#enablereinitialize-boolean
- https://stackoverflow.com/questions/58775482/how-to-reset-form-in-formik-using-enablereinitialize

yedhink (changes requested):
@rishabhanand12 _a

One comment added.

app/javascript/src/components/Dashboard/Settings/Profile.jsx
rishabhanand12: @yedhink Once the form is submitted the response could be one of the following:
- Validation fails for first name or last name. In that case, we don't need to reinitialize.
- Everything goes well and the update is successful. I believe this is the actual use case of `enableReinitialize` since the values have been updated. In this case, we do not need to reinitialize the form because the form will reflect updated values. If user refreshes the page, form will be reinitialized with the latest values.
- Password is incorrect. In that case, we should only reset the password.

Please let me know if I am missing something here.

yedhink: @rishabhanand12 https://www.loom.com/share/330e7fb58bc94a4984cc28b89f9e266b

yedhink:
@rishabhanand12 What's the progress with this PR? Let's try to get this merged with priority.

app/javascript/src/components/Dashboard/Settings/Profile.jsx
rishabhanand12: @yedhink Please take a look.

yedhink: @rishabhanand12 
image

Not sure what you meant.

rishabhanand12: Sorry I forgot to add the link. Here it is https://www.loom.com/share/441a92408315477da67823afe644f37c.

yedhink:
@rishabhanand12 _a

https://www.loom.com/share/8c9af2e23fb9452b8d4f2dbe44a3ba51

yedhink:
@rishabhanand12 Also: https://www.loom.com/share/1d233bb0e4484142ae15a1a5778ce4ba

Use:
- [ ] useMemo
- [ ] Make submit buttons work on pressing enter key


[Edit] More updates on `enableReinitialize` prop and how it updates Formik context: https://www.loom.com/share/b2945f1f68c342f39bedfda61a899d49

rishabhanand12:
@yedhink Please take a look https://vimeo.com/675776245/a13e0a725f

yedhink:
@rishabhanand12 _a

The main issue with overall logic was that we were trying too hard to couple the password confirmation in the same Formik form as the parent. But that was the issue. We were trying to make too many hacky changes to get the job done.

- I have decoupled password confirmation modal and the parent
- This allows for a more finer grain of control over the password confirmation form
- Also solves weird Formik state issues
- The issue with password form not getting submitted on `enter` key was because the form couldn't recognize the deeply nested submit button during first paint. I have fixed that. Refer: https://www.chromium.org/developers/design-documents/create-amazing-password-forms/ and https://github.com/jaredpalmer/formik/issues/1418#issuecomment-524959877
- Made the code as clean and reusable as possible - also removed the custom `equals` logic and also removed the `submitted` state from certain tabs for better UX

I want you to thoroughly test the UI and see if all issues have been fixed and whether any edge cases are still persisting. Also, go through the code too and lemme know if you have suggestions.

Thanks.

app/controllers/api/v1/base_controller.rb
```diff
@@ -1,41 +1,10 @@
...
 
 class Api::V1::BaseController < ApplicationController
-  before_action :authenticate_user_using_x_auth_token
-  before_action :authenticate_user!
+  include ApiResponders
+  include Loggable
+  include ApiRescuable
+  include Authenticable
 
   respond_to :json
```
yedhink, line 9: Do we need this here? I think the routes should handle this.

app/controllers/api/v1/notes_controller.rb
```diff
@@ -2,44 +2,29 @@
 
 class Api::V1::NotesController < Api::V1::BaseController
   before_action :load_note!, only: %i[update delete]
-  before_action :load_notes!, only: :bulk_delete
+  before_action :load_notes, only: :bulk_delete
 
   def index
-    render json: { notes: current_user.notes }
+    render status: :ok, json: { notes: current_user.notes }
```
yedhink, line 8: We can use respond_with_success

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -5,13 +5,12 @@ class Api::V1::SessionsController < Api::V1::BaseController
...
-    user = User.find_for_database_authentication(email: params[:user] && params[:user][:email])
+    user = User.find_for_database_authentication(email: session_params[:email])
     if invalid_password?(user)
-      respond_with_error "Incorrect email or password", 401
+      respond_with_error(t("invalid_credentials"), :unauthorized)
     else
       sign_in(user)
-      render json: { auth_token: user.authentication_token, user: user, is_admin: user.super_admin? },
-        location: root_path, status: :created
+      respond_with_json(:created, { auth_token: user.authentication_token, user: user, is_admin: user.super_admin? })
```
yedhink, line 13: I would prefer to keep it aligned with granite. https://github.com/bigbinary/granite/blob/master/app/controllers/concerns/api_responders.rb#L18-L20

app/controllers/api/v1/users_controller.rb
yedhink: ^ Not sure why above conversation was resolved even when changes weren't made yet. https://github.com/bigbinary/granite/blob/master/app/controllers/concerns/api_responders.rb#L18-L20

app/controllers/concerns/api_rescuable.rb
yedhink: @rishabhanand12 Why is this comment resolved even when changes haven't been made? Why's this `log_exception` still here?

app/controllers/concerns/api_responders.rb
```diff
@@ -0,0 +1,21 @@
...
+      render status: status, json: { error: error_message }.merge(context)
+    end
+
+    def respond_with_success(message, status = :ok, context = {})
+      render status: status, json: { notice: message }.merge(context)
+    end
+
+    def respond_with_json(status = :ok, context = {})
+      render status: status, json: context
+    end
```
yedhink, line 20: Keep it aligned with granite.

app/controllers/concerns/authenticable.rb
```diff
@@ -0,0 +1,24 @@
...
+  included do
+    before_action :authenticate_user_using_x_auth_token
+    before_action :authenticate_user!
+  end
+
+  private
+
+    def authenticate_user_using_x_auth_token
+      user_email = request.headers["X-Auth-Email"]
+      auth_token = request.headers["X-Auth-Token"].presence
```
yedhink, line 15: `presence` can give `nil` value which might make `secure_compare` fail.
```suggestion
      auth_token = request.headers["X-Auth-Token"].to_s   
```

app/javascript/src/components/Dashboard/Settings/utils.js
```diff
@@ -0,0 +1,14 @@
...
+  SETTINGS_NAVLINKS.find(navlink => key === navlink.key);
+
+export const buildProfileFormInitialValues = user => ({
+  firstName: user.first_name,
+  lastName: user.last_name,
+});
+
+export const buildEmailFormInitialValues = user => ({
+  email: user.email,
+  password: "",
```
yedhink, line 13: I think I missed this out. Please remove this key.

app/javascript/src/components/Hero.jsx
```diff
@@ -0,0 +1,35 @@
...
+      </p>
+      <Features features={FEATURE_LIST} />
+      <div className="mt-6 flex items-center justify-center">
+        <Button size="large" type="primary" to="/login" label="Login" />
+      </div>
+    </div>
+  </div>
+);
+
+export default Hero;
```
yedhink, line 35: You can memoize this component.

yedhink (changes requested):
@rishabhanand12 _a

Few comments.

app/javascript/src/components/Hero.jsx
rishabhanand12: @yedhink We can memoize this component but we should be careful here as all code in wheel sends a message about the code practices we use at BigBinary. 

This component will only be rendered once before login. It won't be rendered after the login. As this component is not bulky and not being rendered repeatedly there won't be any significant performance gains.

Please let me know your thoughts on this.

yedhink: That's a fair point. I guess we don't need to memoize then because it's rendered only before login.

rishabhanand12:
@yedhink Please take a look.

app/controllers/concerns/api_responders.rb
rishabhanand12: @yedhink As I mentioned  in https://github.com/bigbinary/bigbinary-website/pull/1931#discussion_r804704875. we should update Granite. 

Please let me know your thoughts.

yedhink:
@rishabhanand12 _a

It was intentional.

- Confirming password means we are checking with the user whether they want to submit the form or not
- If they choose to go ahead with submission, and if some rejection occurs from backend, then we ought to reset the form back to it's original state

So let's keep the same logic intact.

Lemme know if there are any other issues you need to clarify. If none, then please work on the review changes, and let's get this merged ASAP.

rishabhanand12:
@yedhink _a Please review. Please refer to https://vimeo.com/677654720/06c3ef51ba for the changes I have made.

yedhink:
@rishabhanand12 Closing this PR in favor of https://github.com/bigbinary/wheel/pull/782#issue-853202726

Issue #680, opened 2021-10-17
React error when opening the site after force closing the browser


VarunSriram99:
**Steps to recreate**

1. Open a browser and go to the wheel based site.
2. Login to the wheel site
3. Force close the browser(cmd+q in mac) without logging out.
4. Open the site again and it should technically still be logged in but some pages such as edit profile will result in react throwing an error like "cannot get email property of undefined".

**Cause**

The auth-token is validated via react when logging in but the context of ReactJS in wiped during force shutdown. So areas where `import {user} from useUserState` is used, will have the value of user as undefined.

**Possible fix**
Using localStorage to save the context data on login and retrieving data from the localStorage if the context is undefined is a possible fix.

**Code changes**
_app/javascript/src/contexts/user.js_
`const useUserState = () => {
  //const context = React.useContext(UserStateContext);
  let context = React.useContext(UserStateContext);
  if (context === undefined) {
    throw new Error("useUserState must be used within a UserProvider");
  }
  /************New code*******************/
  if (context.user) setToLocalStorage("context", JSON.stringify(context));
  else context = JSON.parse(getFromLocalStorage("context"));
  /********************************************/
  return context;
};`

_app/javascript/src/components/Common/Navbar/index.jsx_
`const handleLogout = async () => {
    try {
      setToLocalStorage("context", null); //Added line of code
      await authenticationApi.logout();
      authDispatch({ type: "LOGOUT" });
      resetAuthTokens();
      window.location.href = "/";
    } catch (error) {
      Toastr.error(error);
    }
  };`

yedhink:
@VarunSriram99 Please format the code properly in your comment with proper syntax highlighting. https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown

rishabhanand12:
> Hi Rishabh, If you see the video, u'll notice that the Avatar has changed. The issue I mentioned affects only the places where useUserContext is used. So if you try to open the edit profile page then you'll see a white screen.

@VarunSriram99 You are right. @yedhink Shall I take this up?

yedhink:
Please test it in both chrome as well as safari. Also what happens when you refresh page?

PR #765, opened 2022-01-21
Updated storage helper methods

### LINKED ISSUE #719: Improve way of managing null values in storage.js
## setToLocalStorage

https://github.com/bigbinary/wheel/blob/ecfddd6aa16d2ae9ec78df160a4f0786f397de6d/app/javascript/src/helpers/storage.js#L1-L3
Should be replaced with:
```js
const setToLocalStorage = (key, value) => {
  if (value != null) {
    localStorage.setItem(key, JSON.stringify(value));
  } else localStorage.removeItem(key);
};
```
Reason: null values need not be saved. The key should be removed in such case.

## getFromLocalStorage

https://github.com/bigbinary/wheel/blob/ecfddd6aa16d2ae9ec78df160a4f0786f397de6d/app/javascript/src/helpers/storage.js#L5-L14
Should be replaced with:
```js
const getFromLocalStorage = key => JSON.parse(localStorage.getItem(key));
```
Reason: When the above change is made, this one-liner will perform the same way as the old one.
Edge cases: when no item exists in localStorage with given key, `localStorage.getItem(key)` retrieves `null`. `JSON.parse(null)` returns `null`

rishabhanand12:
Fixes #719 

@rishabhanand12 _a

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include SetHoneyBadgerContext
+  include Expirable
 
   private
 
--- a/app/controllers/concerns/expirable.rb
+++ b/app/controllers/concerns/expirable.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Expirable
+  extend ActiveSupport::Concern
+
+  included do
+    rescue_from ActionController::InvalidAuthenticityToken, with: :ensure_user_is_logged_out_during_session_expiry
+  end
+
+  private
+
+    def ensure_user_is_logged_out_during_session_expiry
+      respond_with_json({ error: t("session.expiry") }, :unauthorized) and return
+    end
+end
--- a/app/javascript/src/components/Main.jsx
+++ b/app/javascript/src/components/Main.jsx
@@ -18,6 +18,10 @@ import {
 } from "components/routeConstants";
 import { useAuthState, useAuthDispatch } from "contexts/auth";
 import { useUserDispatch, useUserState } from "contexts/user";
+import {
+  clearLocalStorageCredentials,
+  getFromLocalStorage,
+} from "utils/storage";
 
 const Main = props => {
   const [loading, setLoading] = useState(true);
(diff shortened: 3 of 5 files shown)


rishabhanand12:
@yedhink _a Please take a look.

yedhink:
@rishabhanand12 _a

Can you please double verify from UI that this change doesn't affect:
- Login
- Signup
- Logout
- Login and then closing browser and reopening page plus refreshing it

This helper might be used in neeto products too. Thus we need to double verify before merging.

rishabhanand12:
@yedhink I have verified the features as suggested. I have a question. Please take a look https://www.loom.com/share/9d8f7c353d7947d8aed6a0cc972573d8. 

P.S. One thing I forgot to mention in the video is that the behaviour is same in master branch.

yedhink:
@rishabhanand12 _a

This is the issue that lead to use adding that `isLoggedIn` logic in the first place: https://github.com/bigbinary/wheel/issues/680#issue-1028339676

I guess, if user is not logged in, then we should clear the localStorage tokens say in PrivateRoute itself.

> We should use sessions

Why do you think we need to use sessions in this case?

rishabhanand12:
@yedhink We could clear the local state in `PrivateRoute` but that doesn't fix the issue. User should be logged in if the browser is force closed, right?

I suggested to use session storage because I was under the impression that logging the user out after session ends is an intentional feature. In that case it made more sense to use session storage.

yedhink:
Tracking this issue in https://github.com/bigbinary/wheel/issues/774

Putting this PR on hold until we fix that issue.

yedhink:
The issue with browser quit has been fixed.

We use simple auth token based strategy and we persist these tokens in localStorage. But the session of the user is handled via session-cookies, by Devise.

So I think we can add a logic in client side, that checks for presence of `props.user` from backend. There are few cases we need to handle:
- localStorage doesn't have any data - then we don't have to do anything - the current code is good enough
- localStorage has authToken etc but the session has expired - but props.user is NOT present - in this specific case, we should clear the localStorage

@rishabhanand12 _a

Please take this up. Lemme know if you have any queries.

rishabhanand12:
@yedhink _a I have verified that the application is working as intended after #774 fix and I have also checked for cases mentioned in https://github.com/bigbinary/wheel/pull/765#issuecomment-1019878181.

I don't think we need to add any additional logic to clear the localstorage if the session data from devise is null. Please take a look.

PR #767, opened 2022-01-27
Disable logger in production env

yedhink:
@chiraggshah _a

Questions:
- Should we use `RAILS_ENV` over `NODE_ENV` to check for current environment given that there are envs like `heroku`, `staging` etc?
- Should we disable this logger in any other environment, other than just `production`?

Please review and verify.

### DIFF
--- a/app/javascript/src/common/logger.js
+++ b/app/javascript/src/common/logger.js
@@ -2,7 +2,7 @@ export const initializeLogger = () => {
   /* eslint react-hooks/rules-of-hooks: "off" */
   const Logger = require("js-logger");
   Logger.useDefaults();
-  if (process.env.NODE_ENV === "production") {
+  if (process.env.RAILS_ENV === "production") {
     Logger.setLevel(Logger.OFF);
   }
 };
--- a/app/javascript/src/common/logger.js
+++ b/app/javascript/src/common/logger.js
@@ -1,4 +1,8 @@
 export const initializeLogger = () => {
-  /* eslint no-undef: "off"*/
-  require("js-logger").useDefaults();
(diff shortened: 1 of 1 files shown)


chiraggshah:
I think it would be better if we use RAILS_ENV and disable it only for production. That would help us in debugging any errors in staging/heroku environments.

yedhink:
@unnitallman We need to add this to neetoCompliance. Every project needs to have this logger file.

- In development kind of environemnts, this logger helps in debugging.
- In production env this logger is disabled.

In order to achieve both cases, we need this logger initializer file.

But NOT all neeto products follow wheel structure. Example: https://github.com/bigbinary/neeto-desk-web/tree/master/app/javascript

^ There is no `src` folder. Thus we can't sync this logger file given that it's in the path: `app/javascript/src/common/logger.js`.

I can move this file to `app/javascript/logger.js` so that it can be compatible with all neeto project structures. Or should we update all neeto projects to have same structure(atleast the base) as wheel?

Do you think we can take this up with some priority due to it's vulnerabilities? Refer discussion: https://bigbinary.slack.com/archives/C030C59TPPV/p1643267925581649

The logger details can be viewed here(content will be updated soon): https://www.bigbinary.com/learn-rubyonrails-book/setting-up-react-environment#js-logger

cc: @prasanthchaduvula

chiraggshah (approved):
@yedhink _a : Looks good.

unnitallman:
@yedhink 

1. We will go with the structure that Wheel follows.
2. We will add a verifier in neeto-compliance to make sure all neeto apps use the logger.js from wheel. @prasanthchaduvula 

The PR looks good to me.

PR #768, opened 2022-01-27
Added tailwind classes sorter prettier plugin

yedhink:
@karthiknmenon _a

**IMPORTANT: The tailwind config file needs to be in the root of the project as in wheel. Hoping all neeto projects follow the same structure.**

Can you please give this a review from your end and see if the changes that this plugin has made is NOT affecting anything in UI?

Ultimately this plugin only just fixes the order of classNames. Did a verification from my end.

Given that this is a major addition and might show a lot of file changes in neeto side, I would like to get a second verification on this.

cc: @chiraggshah

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  tailwindConfig: "./tailwind.config.js",
   trailingComma: "es5",
   arrowParens: "avoid",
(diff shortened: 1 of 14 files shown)


karthiknmenon (approved):

karthiknmenon:
@yedhink _a LGTM.

PR #789, opened 2022-02-17
Add pry-byebug instead of pry and byebug

### LINKED ISSUE #788: Replace pry and byebug with pry-byebug
We have both pry and byebug gems for debugging.
To make things simple, can we use pry-byebug  :smile:
It supports debugging via both the syntaxes (binding.pry or byebug) and has capabilities of both.

P.S: The author of pry-byebug and byebug is the same. (edited)

chiraggshah:
fixes #788

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,13 +54,12 @@ group :development, :test do
   # Rails integration for factory-bot
   gem "factory_bot_rails"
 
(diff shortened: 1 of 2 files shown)


chiraggshah:
@yedhink _a : Please review

PR #797, opened 2022-03-01
Remove unused momentJS

chiraggshah:
Don't think "moment" is being used directly in the app. Hence removing it.


chiraggshah:
@yedhink _a : Please review.

PR #811, opened 2022-04-01
Sync bin folder to neeto

### LINKED ISSUE #791: Sync files in "bin" folder to all neeto repos
@yedhink _a

yedhink:
Fixes #791 

@unnitallman _a Can you please review and merge this?

This PR ensures that if we upgrade the Rails version in wheel and if any specific changes occurs to the files in `./bin` folder at the root of wheel, then it gets synced to all neeto repos.

In Rails 7 etc, `spring` gem gets removed and thus most of the files under this folder needs to be changed across neeto. 

This PR would automate that part.

### DIFF
--- a/.scripts/sync_with_wheel.sh
+++ b/.scripts/sync_with_wheel.sh
@@ -42,6 +42,15 @@ declare -a configs=(
   ".nvmrc"
   ".ruby-version"
   ".erb-lint.yml"
+  "bin/bundle"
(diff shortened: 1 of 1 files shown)


unnitallman:
@prasanthchaduvula We need to add this to neeto-compliance script along with the current `sync_with_wheel` script. Please create an issue for this in neeto-compliance repo.

PR #818, opened 2022-04-08
Bump rails to 6.1.5

chiraggshah:
Changes: https://github.com/rails/rails/releases/tag/v6.1.5

I believe only ActionCable and ActionMailbox have some major changes which can be looked into. Rest all should be good.


chiraggshah:
@yedhink _a : Please have a look.

PR #835, opened 2022-05-03
Upgrade ruby 3.1.2

chiraggshah:
- Upgraded ruby to 3.1.2
- Used the new hash value omission syntax : https://bugs.ruby-lang.org/issues/14579
- Had to add "net-smtp" gem explicitly as the following gems are not bundled with ruby
```

The following default gems are now bundled gems.

net-ftp 0.1.3
net-imap 0.2.2
net-pop 0.1.1
net-smtp 0.3.1
matrix 0.4.2
prime 0.1.2
debug 1.4.0
```

We used only net-smtp, but if any other applications are using any of the above, those need to be added explicitly to the gemfile.

References: https://github.com/ruby/ruby/blob/v3_1_0/NEWS.md


chiraggshah:
@yedhink _a : Please review. If it looks good, please assign it to Ghouse to add a docker image for 3.1.2.

unnitallman:
Reverting this. 3.1.2 is causing issues with some gems in Neeto eco system.

PR #856, opened 2022-07-09
Fix Husky pre-commit Error

### LINKED ISSUE #853: Fix husky pre-commit error
Please find below the screenshot:-
image

bhavik2936:
Fixes #853

## What
Modified for loop with while loop, which is actively being supported by `sh` shell.

## Why
Looping constructs (i.e. _for_ loop) being called in `.husky/helpers/prevent_conflict_markers.sh:7` is not supported by _sh_ shell as is feature part of _bash_ shell, as mentioned [here](https://www.gnu.org/software/bash/manual/bash.html#Looping-Constructs).

### DIFF
--- a/.husky/helpers/prevent_conflict_markers.sh
+++ b/.husky/helpers/prevent_conflict_markers.sh
@@ -4,7 +4,7 @@ repeat() {
   char="$1"
   times="$2"
   text=""
-  i="0"
+  i=0
 
   while [ $i -lt $times ]
   do
--- a/.husky/helpers/prevent_conflict_markers.sh
+++ b/.husky/helpers/prevent_conflict_markers.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 repeat() {
   char="$1"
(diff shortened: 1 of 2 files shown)


yedhink:
@bhavik2936 Thanks for making the PR. Appreciate it.

But we can't modify the `husky` script because that gets generated automatically by husky. Rather the solution here is to update the Git hook to be `sh` compliant.

Lemme please know if you are planning to work on this requested change. Else I will be closing this PR and taking the issue up by myself in few days.

Thanks.

bhavik2936:
I will try to modify shell iterators to accomodate various shells. By reverting current changes at my earliest feasibility.

bhavik2936:
@yedhink I have accomodated necessary changes, so the PR is ready to review.

.husky/helpers/prevent_conflict_markers.sh
```diff
@@ -4,9 +4,12 @@ repeat() {
   char="$1"
   times="$2"
   text=""
-  for (( i = 0; i < "$times"; ++i ))
+  i="0"
```
yedhink, line 7: Are there any disadvantages of keeping this as an integer itself rather than a string?

bhavik2936: I have modified it to integer value.

bhavik2936:
> The shebang hasn't been updated. Please update it to `#!/usr/bin/env sh`.

I have modified the shebang in `.husky/helpers/prevent_conflict_markers.sh`, let me know if it is the intended location.

PR #862, opened 2022-08-15
Create New Contact #5

naufil50:


naufil50:
Closing

yedhink:
@naufil50 The point of attaching below screenshots is to let you know that you should be extra careful moving forward. If a document has been elaborately created for performing a task, then make sure to read it properly before jumping into implementation.

image

---
image

PR #878, opened 2022-09-12
Remove CODEOWNERS

mr-jestin-roy:


yedhink:
@mr-jestin-roy I would like to know if the following note in the basecamp document concerning the neetoUI challenge is not very clear to you:
image

If it's not very clear, then please do point out what we could mention in the document in order to make it further specific. Thanks.

mr-jestin-roy:
@yedhink Yedhin : Yes,I read that section and it was pretty clear, it was an honest mistake on my part trying to merge branches to my master from VsCode UI itself. Realised it the second it went out and rectified it. Won't repeat the same.

PR #879, opened 2022-09-12
Updated getFromLocalStorage function

### LINKED ISSUE #876: Update `getFromLocalStorage` utils function to handle error
Update the `getFromLocalStorage` function to use try-catch block. So that if some non-parseable value is defined in local storage then it can be caught and an empty string can be returned in that case.
cc: @yedhink

udai1931:
Fixes #876

### DIFF
--- a/app/controllers/concerns/authenticable.rb
+++ b/app/controllers/concerns/authenticable.rb
@@ -15,7 +15,7 @@ module Authenticable
       auth_token = request.headers["X-Auth-Token"].presence
       user = user_email && User.find_by(email: user_email)
 
-      if user && Devise.secure_compare(user.authentication_token, auth_token)
+      if user && auth_token && Devise.secure_compare(user.authentication_token, auth_token)
         sign_in user, store: false
       else
         respond_with_error(t("invalid_credentials"), :unauthorized)
--- a/app/controllers/concerns/authenticable.rb
+++ b/app/controllers/concerns/authenticable.rb
@@ -11,7 +11,7 @@ module Authenticable
   private
 
     def authenticate_user_using_x_auth_token
-      user_email = request.headers["X-Auth-Email"]
+      user_email = request.headers["X-Auth-Email"].presence
       auth_token = request.headers["X-Auth-Token"].presence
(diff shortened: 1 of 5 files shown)


udai1931:
@yedhink _a please check.

yedhink:
@udai1931 _a Update the authenticable concern like I had pointed out in my video. Currently if auth_token is nil, this method has a chance of failure.

https://github.com/bigbinary/wheel/blob/52d1a2e544c987196e1a2a975d5eef0e0852544e/app/controllers/concerns/authenticable.rb#L13-L23

udai1931:
@yedhink _a please check.
> Currently if auth_token is nil, this method has a chance of failure.

I checked this case earlier and it was handled in frontend only but yes there is a chance so also updated the `authenticable` now.

yedhink (changes requested):
@udai1931 _a Shouldn't we be also getting the `presence` object for user email too? Otherwise what happens if the user email key itself is not present in the api request?
```rb
user_email = request.headers["X-Auth-Email"].presence
```

yedhink:
@udai1931 _a If we delete the `authToken` key from localStorage and refresh the page, currently we are not getting any Toastr messages saying it's incorrect email or password. Please send back a relevant message like `Session has expired. Please login once again.` or if there was already a defined error, then use it.

udai1931:
@yedhink https://app.usebubbles.com/3a9iHW7uZD5cVhJmMAZ66Q/toastr-error

yedhink:
@udai1931 _a Thanks for the detailed video. But after taking into consideration the changes required, I think we should ignore this issue of missing toastr when the `authToken` key itself is deleted scenario. We will come back to this in a separate issue. Right now let's focus on the main flaws.

Please revert the latest changes and just keep the PR as it was until [this comment](https://github.com/bigbinary/wheel/pull/879#issuecomment-1243694163) was added.

udai1931:
@yedhink _a reverted the changes, please check.
Also please include me in the right solution's PR/issue for the above problem.

yedhink:
>Also please include me in the right solution's PR/issue for the above problem.

Sure thing.

PR #885, opened 2022-09-14
Fix column types for active storage related tables

chiraggshah:
Faced this issue while cloning a new app with wheel.
All our models have primary key as uuid, but active record attachment was using bigint to store the record id.

Similarly changed it for variants table as well.

Ref: https://edgeguides.rubyonrails.org/active_storage_overview.html#setup

> If you are using UUIDs instead of integers as the primary key on your models you will need to change the column type of active_storage_attachments.record_id and active_storage_variant_records.id in the generated migration accordingly.

### DIFF
--- a/db/migrate/20210519154811_create_active_storage_tables.active_storage.rb
+++ b/db/migrate/20210519154811_create_active_storage_tables.active_storage.rb
@@ -18,7 +18,7 @@ class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
 
     create_table :active_storage_attachments do |t|
       t.string :name, null: false
-      t.references :record, null: false, polymorphic: true, index: false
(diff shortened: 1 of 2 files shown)


chiraggshah:
@yedhink _a : Please review

yedhink:
@chiraggshah Shouldn't we be creating a new migration for the same? Is it fine to modify the existing migration in this case?

chiraggshah:
@yedhink : Ideally yes, we shouldn't be modifying existing old migrations. 
But wheel is a boilerplate/base application and no one is using wheel in production. People would clone it and start afresh, and migrations would always be run on a new/fresh database, that's why I modified the existing migration instead of adding a new file to the boilerplate.

PR #890, opened 2022-09-15
7 create new note UI

rohit-vishwakarma:


rohit-vishwakarma:
Sorry, Yedhin this will never happen again. From now I will be more careful about making PR against the correct repo.

PR #894, opened 2022-09-15
Upgraded neetoUI to v4

### LINKED ISSUE #891: Upgrade neetoUI from v3 to v4
@udai1931 _a Please take this up with priority.

udai1931:
Fixes #891
@udai1931 _a


app/javascript/src/components/Dashboard/Notes/index.jsx
```diff
@@ -66,11 +67,13 @@ const Notes = () => {
                 disabled={!selectedNoteIds.length}
                 icon={Delete}
                 label="Delete"
+                size="small"
                 onClick={() => setShowDeleteAlert(true)}
               />
             }
           />
           <Table
+            rowSelection
```
udai1931, line 76: @yedhink checkbox not working in the table even after adding `rowSelection`. Also  `rowSelection` is not mentioned in docs its just mentioned in a example below in table section. 
Ref: https://neetoui.onrender.com/iframe.html?viewMode=docs&id=components-table--table-without-checkbox#table-without-checkbox

app/javascript/src/components/Dashboard/Settings/ConfirmPasswordFormModal.jsx
```diff
@@ -51,14 +51,12 @@ const ConfirmPasswordFormModal = ({
...
                 type="submit"
                 onClick={handleSubmit}
               />
               <Button
                 disabled={isSubmitting}
                 label="Cancel"
-                size="large"
                 style="text"
                 onClick={onClose}
               />
```
yedhink, line 62: Can you raise an issue and update the https://academy.bigbinary.com/learn-react/pages/formik-forms#:~:text=Formik%20cancel%20button%20type section with this scenario where we shouldn't pass in the `type` as `reset`? The reason is that, we are invoking the reset separately outside of the JSX.

Go through this component to understand.

package.json
```diff
@@ -61,8 +61,10 @@
   "dependencies": {
     "@babel/preset-react": "^7.16.7",
     "@bigbinary/neeto-icons": "^1.8.17",
-    "@bigbinary/neetoui": "^3.2.29",
+    "@bigbinary/neetoui": "^4.0.3",
+    "@emotion/is-prop-valid": "^1.2.0",
     "@rails/webpacker": "5.4.3",
+    "antd": "^4.23.1",
```
yedhink, line 67: Just to confirm. Does neetoUI v4 require this package as whole?

yedhink (changes requested):
@udai1931 _a

app/javascript/src/components/Dashboard/Notes/Pane/Form.jsx
```diff
@@ -54,19 +54,13 @@ const NoteForm = ({ onClose, refetch, note, isEdit }) => {
...
               type="submit"
               onClick={() => setSubmitted(true)}
             />
-            <Button
-              label="Cancel"
-              size="large"
-              style="text"
-              onClick={onClose}
-            />
+            <Button label="Cancel" style="text" onClick={onClose} />
```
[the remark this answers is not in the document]
udai1931: @yedhink Added the reset type but I don't think reset type it's required here. because cancel button closes the pane and every time pane is opened it will use formik initial values.

app/javascript/src/components/Dashboard/Settings/ConfirmPasswordFormModal.jsx
udai1931: I have added a line about this but yes a proper example will be better.

package.json
udai1931: From compilation error I checked we are using antd/lib/date-picker and antd/lib/table but how to install them specifically?

app/javascript/src/components/Dashboard/Notes/index.jsx
yedhink: As discussed in the call, check the neetoUI table and raise an issue there if need be.

udai1931:
@yedhink _a please check.

PR #898, opened 2022-09-19
Cleaned up the base

nitinprojects:


yedhink:
@nitinprojects Can you please point out which part in the following step from the neetoUI challenge is not very clear to you? This way we can update that step in the docs to ensure folks don't make the PR against the wrong repo/branch like you have done right now.

From my perspective that step is pretty clear. But it may not be to someone else. Thus the question.

image

PR #903, opened 2022-09-21
Allowed the usage of json and mp3 extensions in import

josephmathew900:
Fixes #902

PR #904, opened 2022-09-21
Fix import resolvers in eslint

yedhink:
@josephmathew900 _a Please verify the Eslint config changes from this PR. The other changes are side effects of applying the eslint configs.

With https://github.com/bigbinary/wheel/pull/903 I was able to import relative paths with extension.

Example:
- If we add a `.jsx` extension for `Main` in [this import](https://github.com/bigbinary/wheel/blob/master/app/javascript/src/App.jsx#L6), then the eslint rule would NOT complain.

Our expectation is that it will complain like so, which I have resolved in this PR:
image

Can you please check out how this config is playing in few of the neeto apps and if everything looks good, merge this PR? If there are any issues then lemme know.

### DIFF
--- a/.eslint-rules/imports/enforced.js
+++ b/.eslint-rules/imports/enforced.js
@@ -22,7 +22,7 @@ module.exports = {
       "never",
       {
         ignorePackages: true,
-        pattern: { json: "always", mp3: "always" }
+        pattern: { json: "always", mp3: "always" },
       },
     ],
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,12 @@ module.exports = {
     react: {
       version: "detect",
     },
+    // We need this for the import/extensions rule to work: https://github.com/import-js/eslint-plugin-import#importextensions
+    "import/resolver": {
+      node: {
+        extensions: [".js", ".jsx", ".ts", ".tsx", ".svg", ".json", ".mp3"],
+      },
+    },
   },
   parserOptions: {
     ecmaFeatures: {
--- a/app/javascript/src/components/Dashboard/Settings/constants.js
+++ b/app/javascript/src/components/Dashboard/Settings/constants.js
@@ -1,33 +1,5 @@
 import * as yup from "yup";
 
-import Email from "./Email";
-import Password from "./Password";
-import Profile from "./Profile";
-
-export const SETTINGS_NAVLINKS = [
-  {
-    key: "profile",
-    label: "Profile",
-    description: "Manage user",
-    path: "/settings?tab=profile",
-    component: Profile,
-  },
-  {
-    key: "email",
-    label: "Email",
-    description: "Manage email",
-    path: "/settings?tab=email",
-    component: Email,
-  },
-  {
-    key: "password",
-    label: "Password",
-    description: "Manage password",
-    path: "/settings?tab=password",
-    component: Password,
-  },
-];
-
 export const PROFILE_FORM_VALIDATION_SCHEMA = yup.object().shape({
   firstName: yup.string().required("Required"),
   lastName: yup.string().required("Required"),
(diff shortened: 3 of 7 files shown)


yedhink:
@josephmathew900 

> It throws error when I give .jsx but not when I give .js

- That's because it can't resolve a module with the `.js` extension. Even if eslint doesn't throw error our webpack compilation should throw an error. If webpack is throwing an error then I think there won't be any issues. Lemme know if that resolves this particular case or if there's more to it.

> cyclic dependency

- We can make the cyclic dependency rule from eslint as a warning for neeto repos. 
- But it's definitely a bad practice. 
- We had faced a lot of issues in our compilation processes long back due to the same issue. Not sure whether it's relevant right now. But still I think it's a practice that should be avoided. We may not be able to rectify this issue in all repos. But for code that's getting added in future, devs should be warned about the same.

yedhink:
@josephmathew900 _a

>Now I just tried out this configuration in neetoChat and I think ignorePackages is not working as expected.
Unexpected use of file extension "js" for "emoji-regex/RGI_Emoji.js"  

I believe we don't need to pass `.js` for resolving this file. If we remove `.js` then is this eslint rule throwing any errors? Also for experimenting:
- Add the exact same config as in this PR to neetoChat
- Create another file within the `emoji-regex/` folder named something like `xyz.svg`.
- Then import that file.
- If eslint throws an error saying that the ".svg" extension is not needed, then make a call on whether `svg` extension ought t be mentioned for an `svg` import or not. Make a call as in set a standard across the neeto ecosystem.
- If we need `svg` extension to be specified during import, then add it to `pattern` key with an `always`.

In general the resolved extensions mentioned in our settings, that is `extensions: [".js", ".jsx", ".ts", ".tsx", ".svg", ".json", ".mp3"]`, needn't be imported with those extensions. We should avoid the extension. 

Lemme know your thoughts.

cc: @Amaljith-K

josephmathew900:
@yedhink As per the inputs you have given in the above comment, I checked out how this config is playing in few host applications and everything is working perfectly except when we have files with same name but different extension as show in this [example](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md#exception). But I think this is rare case and if it occurs we can add the extension and ignore this rule for that particular import statement. 

Also I think instead of adding `json` and `mp3` in the [pattern](https://github.com/bigbinary/wheel/blob/fb44f4f583d0fa4de9d69f853c33c4f963a1aa38/.eslint-rules/imports/enforced.js#L25) key  we should add these extensions in the [webpacker.yml](https://github.com/bigbinary/wheel/blob/fb44f4f583d0fa4de9d69f853c33c4f963a1aa38/config/webpacker.yml#L36) file under `extensions`.

Lemme know your thoughts.

yedhink:
>Also I think instead of adding json and mp3 in the [pattern](https://github.com/bigbinary/wheel/blob/fb44f4f583d0fa4de9d69f853c33c4f963a1aa38/.eslint-rules/imports/enforced.js#L25) key we should add these extensions in the [webpacker.yml](https://github.com/bigbinary/wheel/blob/fb44f4f583d0fa4de9d69f853c33c4f963a1aa38/config/webpacker.yml#L36) file under extensions.

@josephmathew900 I am not sure how ESLint can read the configs from `webpacker.yml`. Can you point out why exactly you are suggesting to remove the `pattern` key and add the same to webpacker?

josephmathew900:
@yedhink  https://www.loom.com/share/046e0cfb23894f5a914264adf6a20d2a

yedhink:
> @yedhink https://www.loom.com/share/046e0cfb23894f5a914264adf6a20d2a

We should force the developer to always suffix with uncommon extensions like `mp3`, `json` etc. Like how often do we import such files as modules into JS? Rarely from my understanding. In such cases, to provide clarity, we should force the devs to always add the file extension for such rare extensions.

`js,jsx`, `ts,tsx`, `svg`, `png` etc are the kind of extensions that we often need to import as a JS module. Other extensions I believe should be resolved only if there is a specific extension specified for them. Makes the code more explicit.

Lemme know your thoughts regarding the same.

josephmathew900:
@yedhink Sounds good. Please resolve the conflict and merge the PR. I will add this to `neeto_commons`.

PR #921, opened 2022-11-07
Added script and new extensions

### LINKED ISSUE #920: Recommend new extensions for VSCode
image

Although in the book we are asking to install Prettier separately, the Prettier extension is ignored on purpose in Wheel. Have faced a lot of false positives when combined with ESLint. A beginner might go down the rabbit hole of trying to fix the `red squiggly` lines that the Prettier extension throws, which might be a false positive. Source of truth should be the command line and all prettier formatting will be done either ways via Git hooks.

@udai1931 _a Add these extensions to https://github.com/bigbinary/wheel/blob/main/.vscode/extensions.json

udai1931:
Fixes #920, Part of https://github.com/bigbinary/bigbinary-website/issues/2520

### DIFF
--- a/.scripts/setup_vscode.sh
+++ b/.scripts/setup_vscode.sh
@@ -3,7 +3,6 @@
 require "json"
 
 config = `curl -s 'https://raw.githubusercontent.com/bigbinary/wheel/main/.vscode/extensions.json'`
-# echo $extensions
 extensions = JSON.parse(config)["recommendations"]
 
 extensions.each do |extension|
--- a/.scripts/setup_vscode.sh
+++ b/.scripts/setup_vscode.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env ruby
+
+require "json"
+
+config = `curl -s 'https://raw.githubusercontent.com/bigbinary/wheel/main/.vscode/extensions.json'`
+# echo $extensions
+extensions = JSON.parse(config)["recommendations"]
+
+extensions.each do |extension|
+  output = `code --install-extension #{extension}`
+  puts output
+end
--- a/.scripts/setup_vscode.sh
+++ b/.scripts/setup_vscode.sh
@@ -2,10 +2,21 @@
 
 require "json"
 
-code_file_location = `which codes`
-code_insiders_file_location = `which code-insiders`
+def which(cmd)
+  exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+  ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
+    exts.each do |ext|
+      exe = File.join(path, "#{cmd}#{ext}")
+      return exe if File.executable?(exe) && !File.directory?(exe)
+    end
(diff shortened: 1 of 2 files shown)


udai1931:
@yedhink _a please check.

.scripts/setup_vscode.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
+
+require "json"
+
+config = `curl -s 'https://raw.githubusercontent.com/bigbinary/wheel/main/.vscode/extensions.json'`
+extensions = JSON.parse(config)["recommendations"]
+
+extensions.each do |extension|
+  output = `code --install-extension #{extension}`
```
yedhink, line 9: We need to check whether `code` or `code-insiders` is a valid executable command, as is it's available under the `PATH` or not. Then only we should continue with this statement.

If either `code` or `code-insiders` is missing, then print the message `Couldn't find an executable for VSCode. Please ensure that the code or code-insiders command is available in your PATH env`.

yedhink (changes requested):
@udai1931 _a Can you please add a demo video for the same? Want to see what happens if certain extensions are already installed and how new extensions are getting installed. Thanks.

udai1931:
@yedhink _a Demo and commands check: https://app.usebubbles.com/6CbjSF1JWd52jEG8a9pw9f/my-video-recording

.scripts/setup_vscode.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+
+require "json"
+
+code_file_location = `which codes`
```
yedhink, line 5: Had you actually ran this script? Because here it should be `code`.

.scripts/setup_vscode.sh
```diff
@@ -0,0 +1,19 @@
...
+code_insiders_file_location = `which code-insiders`
```
yedhink, line 6: Don't use `which`. Refer: https://stackoverflow.com/a/677212

Use: https://stackoverflow.com/a/5471032

.scripts/setup_vscode.sh
```diff
@@ -0,0 +1,19 @@
...
+if code_file_location.empty? && code_insiders_file_location.empty?
+  puts "Couldn't find an executable for VSCode. Please ensure that the code or code-insiders command is available in your PATH env"
+  return
+end
+
+config = `curl -s 'https://raw.githubusercontent.com/bigbinary/wheel/main/.vscode/extensions.json'`
+extensions = JSON.parse(config)["recommendations"]
+
+extensions.each do |extension|
+  output = code_file_location.empty? ? `code-insiders --install-extension #{extension}` : `code --install-extension #{extension}`
```
yedhink, line 17: Create the executable outside of this loop so that you can avoid performing this condition each and every time.

yedhink (changes requested):
@udai1931 _a

udai1931:
@yedhink _a please check.

PR #930, opened 2022-12-05
Cleaned the project base

ubaidchalil:
- Removed the CODEOWNERS file

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-*       @yedhink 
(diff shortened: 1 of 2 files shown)


yedhink:
@ubaidchalil Can you please point out whether the following statement(attached as screenshot) that we had written in **bold** in the Basecamp doc is clear to you or not? If it's not very clear, please suggest how we can rephrase the sentence so that in future we can avoid PRs and unnecessary notifications like this one. Thanks.

image

ubaidchalil:
@yedhink The message is very clear. The mistake was from my side. I created the PR accidentally.

PR #954, opened 2023-02-06
Cleaned up base app

jenipharachel:
Fixes #1 

@yedhink _a Please review

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-*       @yedhink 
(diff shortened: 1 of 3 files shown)


jenipharachel:
Closing this PR as it was created in wrong repo

yedhink:
@jenipharachel Moving forward, I would recommend that you pay very close attention to the documentation and requirements to avoid mistakes like the one made over here. If you feel that it's an issue with the documentation, then by all means, please let me know what we can do in order to improve the same.

image

image

PR #967, opened 2023-04-25
Update CI and bump ruby version (patch)

abrahamparayil:
- Bump ruby version fom `3.1.2` to `3.1.3` in CI and Dockerfile.
- Syncs CI steps with our current setup that we use everywhere.
- @abrahamparayil _a


abrahamparayil:
@yedhink _a Please review.

PR #974, opened 2023-06-07
Added cypress neetoCI config

jagannathBhat:


yedhink:
@jagannathBhat _a Do we need this in wheel? Unless this file needs to be curled to neeto apps, I don't see a good reason to keep this config in Wheel. I understand there are cypress related files in Wheel, but we don't do any active cypress tests here.

jagannathBhat:
@yedhink Please keep this pull request for now. We are just testing it.

jagannathBhat:
I will close the pull request by EOD

PR #1003, opened 2023-09-11
Replaced rebornix.ruby with Shopify.ruby-lsp

### LINKED ISSUE #1002: Update Ruby extension in VSCode
https://github.com/bigbinary/wheel/blob/c118239933b64a377e4e9c618bc747a38d8640bc/.vscode/extensions.json#L5

^ Please do a global search to find other places where necessary changes ought to be made. We have to update the .vscode folder of `granite` repo similarly.

@navaneethsdk _a We need to remove this extension as it has been deprecated in favor of https://github.com/Shopify/vscode-ruby-lsp.  The transition is smooth - we just need to uninstall current extension and the install the new one and then add the following to the end of the User Settings(JSON) file:
```json
  "[ruby]": {
    "editor.defaultFormatter": "Shopify.ruby-lsp",
    "editor.formatOnSave": true,
    "editor.formatOnType": true,
    "editor.insertSpaces": true,
    "files.insertFinalNewline": true,
    "editor.semanticHighlighting.enabled": true
  },
```

Make necessary changes in LRRB as well as neetoEngineering KB articles.

PS: If you already have some settings in your local `.vscode` folder within the project, then that ought to be updated with above mentioned settings.

navaneethsdk:
Fixes #1002

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,21 +1,15 @@
 {
   "[ruby]": {
     "editor.autoClosingBrackets": "beforeWhitespace",
+    "editor.formatOnType": true,
+    "editor.insertSpaces": true,
+    "files.insertFinalNewline": true,
+    "editor.semanticHighlighting.enabled": true,
     "editor.formatOnSave": true,
-    "editor.defaultFormatter": "rebornix.ruby"
+    "editor.defaultFormatter": "Shopify.ruby-lsp"
   },
   "files.trimTrailingWhitespace": true,
   "files.trimFinalNewlines": true,
   "editor.formatOnSave": true,
   "yaml.format.enable": true,
-  "ruby.useBundler": true, //run non-lint commands with bundle exec
-  "ruby.useLanguageServer": true, // use the internal language server (see below)
-  "ruby.lint": {
-    "rubocop": {
-      "useBundler": true,
-      "lint": true, // enable lint cops
-      "rails": true // requires rubocop-rails gem for RuboCop >= 0.72.0
-    }
-  },
-  "ruby.format": "rubocop",
 }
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,21 +1,15 @@
...
-  "ruby.useBundler": true, //run non-lint commands with bundle exec
-  "ruby.useLanguageServer": true, // use the internal language server (see below)
-  "ruby.lint": {
-    "rubocop": {
-      "useBundler": true,
-      "lint": true, // enable lint cops
-      "rails": true // requires rubocop-rails gem for RuboCop >= 0.72.0
-    }
-  },
-  "ruby.format": "rubocop",
```
navaneethsdk, line 20: Should I be looking for a similar configuration for `ruby-lsp`

.vscode/settings.json
```diff
@@ -1,21 +1,15 @@
 {
   "[ruby]": {
     "editor.autoClosingBrackets": "beforeWhitespace",
+    "editor.formatOnType": true,
+    "editor.insertSpaces": true,
+    "files.insertFinalNewline": true,
```
yedhink, line 6: Let's remove this particular setting. This contradicts with `trimFinalNewlines` I think.

navaneethsdk (commented):
@yedhink Kindly review.

yedhink (changes requested):
@navaneethsdk _a Please check the comment.

.vscode/settings.json
yedhink: Not right now. We can do that as next step. Right now lets get the bare essentials out to public.

navaneethsdk:
@yedhink I have made the said changes. Please see.

PR #1005, opened 2023-09-20
Updated neetoCI workflows

jagannathBhat:
Part of bigbinary/neeto-ci-web#1128


jagannathBhat:
@yedhink _a Please review

PR #1012, opened 2023-10-24
Removes response variable from getFromLocalStorage

### LINKED ISSUE #998: Remove `response` variable from `getFromLocalStorage`
We can change to 
```
const getFromLocalStorage = key => {
  try {
    const value = localStorage.getItem(key);

    return isPresent(value) ? JSON.parse(value) : null;
  } catch (error) {
    logger.error(error);
    return null;
  }
};
```

navaneethsdk:
Fixes #998


navaneethsdk:
@yedhink _a Please review

Issue #1008, opened 2023-09-21
The image for no data is broken


AbhayVAshokan:
![image](https://github.com/bigbinary/wheel/assets/35297280/1eac76b1-9fef-4889-b2c4-7f84482b4fb3)

We have already moved away from the image approach. So for consistency, please delete the image instead.

PR #1013, opened 2023-10-24
Incorporates neeto-ui NoData convention for EmptyState component

navaneethsdk:
Fiixes #1008


navaneethsdk:
@yedhink _a Please review.

Reference: https://github.com/bigbinary/neeto-ui/blob/3b4ae6c3c20fd4ed07bea2edec9e893edeae2197/src/components/NoData/index.jsx#L37

